Issue #5, opened 2020-08-26
Finalize partition layout


tannewt:
Hey, could we finalize the partition layout? I've added native WiFi to the CircuitPython build and I'm over `0x130000` bytes per partition. I'm thinking we'll want to do 1MB for the filesystem and then split the rest between two OTA and UF2. We aren't using NVS at all. What size do we want to leave for UF2?

It'd be great to share the partition layout with Arduino as well. @me-no-dev what partition layout are you using for Arduino?

hathach:
currently the bootloader take 200 KB, I will try to see if I could optimize it, it seems like lto is not enabled for IDF. I will try to see if I could shrink it down furthermore. 

```
            Archive File DRAM .data & .bss & other   IRAM   D/IRAM Flash code & rodata   Total
                  libc.a        364      4       0      0      368      54973     3794   59503
                   (exe)         92   5281       0      0     5376      20749      476   31974
               libmain.a        417  10697       0      0    11114       1869     1091   25188
           libfreertos.a       2080    736       0      0    12663        133     1386   16998
                libsoc.a       1058      0       0      0     9260       3305      947   14570
             libdriver.a        124     66       0      0     1780       7362     2137   11469
          libspi_flash.a       1053    416       0      0     7245       1358     1228   11300
         libesp_system.a       2363     21       0      0     5806        785      792    9767
         libesp_common.a        153     84       0      0      452       1375     5862    7926
            libesp32s2.a         16     21       0      0     1246       2979     1537    5799
                libvfs.a        272     48       0      0      320       4389      560    5589
               libheap.a        301      8       0      0     3003        830      729    4871
 libbootloader_support.a          0      4       0      0        4       2717     1715    4440
        libesp_ringbuf.a          0      0       0      0     2934          0      411    3345
             libnewlib.a        190    264       0      0     1367        724       91    2636
         libmbedcrypto.a          0      0       0      0        0       1459      365    1824
         libapp_update.a          1     12       0      0      167        612      608    1400
                liblog.a          8    272       0      0      486        383      150    1299
          libesp_timer.a         16     20       0      0      332        607      226    1201
              libefuse.a         28      4       0      0       32        720      394    1178
          libled_strip.a          0     16       0      0      181        597      377    1171
             libxtensa.a          0      0       0      0      409          0       58     467
             libboards.a          0      9       0      0        9        367       20     405
                libhal.a          0      0       0      0      333          0       32     365
[...]

PR #8, opened 2020-09-09
finialize partition table

hathach:
implement #5


tannewt (approved):
Thank you! Want to create a default UF2 file to load it with?

hathach:
@tannewt I will update the github action to upload bin file when release, we can use that as part of circuitpython upload

PR #10, opened 2020-09-23
Support multiple flash sizes

tannewt:
And also boards without NeoPixels.


tannewt:
cc @unexpectedmaker

UnexpectedMaker:
Awesome! So if I wanted to add my ProS2 as well, I can just duplicate the feathers2 stuff?

tannewt:
Yup! Just another board def.

components/boards/espressif_saola_1_wrover/board.h
```diff
@@ -45,7 +45,7 @@
 // GPIO connected to Neopixel data
 // Note: On the production version Saola (v1.2) it is GPIO 18,
 // however on earlier revision v1.1 it is GPIO 17
-#define PIN_NEOPIXEL          18
+#define PIN_NEOPIXEL          17
```
hathach, line 48: production run saola board has neopixel on pin 18, you should keep this as local modification.

components/boards/espressif_saola_1_wrover/board.h
```diff
@@ -36,7 +36,7 @@
 
 // GPIO that implement 1-bit memory with RC components which hold the
 // pin value long enough for double reset detection.
-// #define PIN_DOUBLE_RESET_RC
+#define PIN_DOUBLE_RESET_RC 13
```
hathach, line 39: saola doesn't have this feature implementation, I think we should leave this as local modification for testing.

components/boards/espressif_saola_1_wrover/board.h
```diff
@@ -56,7 +56,7 @@
 
 // LED for indicator and writing flash
 // If not defined neopixel will be use for flash writing instead
-#define PIN_LED               33
+#define PIN_LED               36
```
hathach, line 59: saola doesn't has regular led, this should be commented out. It is my bad somehow I got it committed to the code base. I think I intended to test with it previously, but ended up pulling to other work.

hathach (commented):
great PR, though for saola board modifictaion, I think we should keep `PIN_DOUBLE_RESET_RC` and `PIN_NEOPIXEL` as local modification for testing.

tannewt (commented):
Yup! Agreed. I removed and rebased so the history wouldn't impact that file.

hathach (approved):
it looks great now :)

PR #11, opened 2020-09-26
Upload release assets and build artifacts

tannewt:
Also, add UnexpectedMaker FeatherS2


hathach (approved):
superb !! Thank you for the PR, I was trying to do something similar, but got stuck with daplink v2.

Issue #12, opened 2020-09-27
Hard Lock-up on Saola-Wroom


julianrendell:
Up front: this is probably user error...

I've built and flashed the latest from this repo's master branch.

I am using the latest idf-tools. 

I've enabled level 2 TinyUSB logging and added some extra logging:

```
--- idf_monitor on /dev/cu.usbserial-14430 115200 ---
--- Quit: Ctrl+] | Menu: Ctrl+T | Help: Ctrl+T followed by Ctrl+H ---
ESP-ROM:esp32s2-rc4-20191025
Build:Oct 25 2019
rst:0x1 (POWERON),boot:0x8 (SPI_FAST_FLASH_BOOT)
SPIWP:0xee
mode:DIO, clock div:1
load:0x3ffe6100,len:0x8
load:0x3ffe6108,len:0x1aa8
load:0x4004c000,len:0x1cd4
load:0x40050000,len:0x1c08
entry 0x4004c454
I (46) boot: ESP-IDF v4.3-dev-1197-g8bc19ba89 2nd stage bootloader
I (46) boot: compile time 19:13:47
I (46) boot: chip revision: 0
I (49) boot.esp32s2: SPI Speed      : 80MHz
I (54) boot.esp32s2: SPI Mode       : DIO
I (59) boot.esp32s2: SPI Flash Size : 4MB
I (63) boot: Enabling RNG early entropy source...
I (69) boot: Partition Table:
I (72) boot: ## Label            Usage          Type ST Offset   Length
I (80) boot:  0 nvs              WiFi data        01 02 00009000 00005000
I (87) boot:  1 otadata          OTA data         01 00 0000e000 00002000
I (95) boot:  2 ota_0            OTA app          00 10 00010000 00160000
I (102) boot:  3 ota_1            OTA app          00 11 00170000 00160000
I (110) boot:  4 uf2              factory app      00 00 002d0000 00040000
I (117) boot:  5 user_fs          Unknown data     01 81 00310000 000f0000
I (125) boot: End of partition table
I (129) boot: Defaulting to factory image
I (134) boot: JULIAN- boot_index: -1
I (138) boot: JULIAN- RESET_REASON: 1
I (142) boot: JULIAN- boot_index-2: -1
I (147) esp_image: segment 0: paddr=0x002d0020 vaddr=0x3f000020 size=0x037f4 ( 14324) map
[...]

PR #13, opened 2020-09-29
Clarifying need for patched esp-idf

julianrendell:


julianrendell:
This is related to #12 .

PR #17, opened 2020-10-30
Updated readme to list the FeatherS2 as a supported board 

UnexpectedMaker:
and to indicate the FeatherS2 requires the "RESET" then "BOOT" sequence to get it into the UF2 bootloader.


README.md
```diff
@@ -13,7 +14,7 @@ There are a few ways to enter UF2 mode:
 - There is no `ota application` and/or `ota_data` partition is corrupted
 - `PIN_BUTTON_UF2` is gnd when 2nd stage bootloader indicator is on e.g **RGB led = Purple**. Note: since most ESP32S2 board implement `GPIO0` as button for 1st stage ROM bootloader, it can be used for dual-purpose button here as well. The difference is the pressing order:
   - Holding `GPIO0` then reset -> ROM bootloader
-  - Press reset, see indicator on (purple RGB) then press `GPIO0` -> UF2 bootloader
+  - Press reset, see indicator on (purple RGB) then press `GPIO0` or the boot button -> UF2 bootloader (required on the UM FeatherS2)
```
hathach, line 17: this is the same for many boards, this readme is mostly for developer guide and just explain how the feature works. I don't think we could list all the boards that use this method, the boot button can be name DFU button on other board. Therefore I think we just leave it as it is. Specific board probably need its own specific user guide for end-user.

hathach (changes requested):
there is a  massive pending changes on the way that makes this repo cross-platform
https://github.com/adafruit/uf2-esp32s/pull/16
the readme got relcoated, would you mind if we leave this pending until above PR is merged then you could rebase this PR.

PR #32, opened 2020-11-24
Feature load2ram

nxf58843:
Work in progress.  Trying to get imxrt1010_evk to target RAM so that it can be loaded with ROM serial downloader.
It builds but does not connect when loaded.

### DIFF
--- a/ports/mimxrt10xx/boards/imxrt1010_evk/flash_config.c
+++ b/ports/mimxrt10xx/boards/imxrt1010_evk/flash_config.c
@@ -10,28 +10,13 @@
 #include "boards.h"
 
 
-__attribute__((section(".boot_hdr.sdp")))
-/*************************************
- *  IVT Data
- *************************************/
-const ivt sdp_ivt = {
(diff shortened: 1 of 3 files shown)


hathach:
you shouldn't modify the current working flash linker on 1010_evk. Instead you should add an extra board called `imxrt1010_evk_ram` that will allow user to pick either flash/ram at their choice.

nxf58843:
The eventual goal will be to generate a load to RAM image that can be copied from flash by the internal ROM, or loaded through the serial download protocol.  We should be able to support both with the same build.

hathach:
> The eventual goal will be to generate a load to RAM image that can be copied from flash by the internal ROM, or loaded through the serial download protocol. We should be able to support both with the same build.

ah thanks, to be honest, I am a bit clueless on this load2ram :D . Hopefully others who have done something similar could join in and help

nxf58843:
Made progress in another branch.  closing this request.

hathach:
No problems at all, feel free to make the new pr whenever you think it is ready

PR #40, opened 2020-12-18
RT1011 Link2ram

nxf58843:
This has basic functionality with imxrt1010_evk
I was able to load this image into the RT1010 EVK using sdphost.exe and then update the bootloader with a UF2 file created from the same binary.
Instructions for loading and updating are in ports/mimxrt10xx/load2ram.md


nxf58843:
More functional update coming

PR #41, opened 2020-12-21
Link2ram

nxf58843:
This updates the code to run from RAM.  

The headers are generated so that the image can be copied directly to flash and get loaded into RAM by the boot ROM.

This same image can be loaded directly into RAM with the serial download protocol.  

The imxrt1010_evk and imxrt1060_evk have been tested.  The metro_m7_1011 builds, but has not been tested.

See ports/mimxrt10xx/load2ram.md for more details.


hathach:
@nxf58843 thank you very much for your PR, sorry for late response. I am on vacation the last week. I am reviewing and testing out this PR, will provide the feedback as soon as I could.

hathach:
@nxf58843 thank you very much for the load2ram configuration. From what have been testing with this PR, since SDP load image to ram, we have to actually flash the bootloader on qspi afterwads for it to survive a reset. It is a nice feature to easily recover a bricked board without using external debugger. However, since it is multiple steps involved
1. Set BootMode to 01, use sdphost to load and execute it from ram
2. Then program qspi flash with bootloader
3. Set the BOOT=10 for normal execution

For production run, this take time since it require manual interaction by people. I would prefer to just simply run `pyocd flash` with an daplink/jlink to program the qspi flash and have it ready to go.

If this is the case, I think we should separate the ram config from normal board that is designed to used with debugger such as  imxrt1010_evk which is normally flashed using pyocd e.g `imxrt1010_evk_ram`  where you could specify its own linker script (memory section), flash_config.c etc ... without affecting normal config `imxrt1010_evk` for nor flash. Since I am not familiar with this feature, maybe I didn't get it right, any feedback is welcome.

nxf58843:
@hathach the generated binary can be loaded directly into flash with pyocd.  I use the same binary to create a .uf2 file.  You just can't load the .hex file into flash because the addresses are referenced to RAM.

Also, when the flash is blank, boot mode 10 will fail through to serial download mode so the mode would not need to be changed if the flash is blank, only if it is corrupted for recovery.

We can streamline the process further using the bootloader revision.  We can add code to check the image revision in flash, and if the image in RAM is newer, it can copy itself to flash.  Is the new revision located at a specific address that will not change between updates?  If not, can we put it at a fixed address?

This image can be loaded into flash or RAM and can be loaded without additional debugger hardware.  If we implement the automated revision check (or blank check), it should be faster to load through USB than through SWD.

hathach:
sorry for late response, still catching up with works after the vacation.
 
> @hathach the generated binary can be loaded directly into flash with pyocd. I use the same binary to create a .uf2 file. You just can't load the .hex file into flash because the addresses are referenced to RAM.

Could you please tell me which format/command that we could use the pycod with to flash the bootloader to qspiflash.

> Also, when the flash is blank, boot mode 10 will fail through to serial download mode so the mode would not need to be changed if the flash is blank, only if it is corrupted for recovery.

That is great to hear, I just want to confirm that after the bootloader is flashed, and bootmode = normal. The uf2 bootloader will run without the need of `sdhost` to load it to SRAM right ?

> We can streamline the process further using the bootloader revision. We can add code to check the image revision in flash, and if the image in RAM is newer, it can copy itself to flash. Is the new revision located at a specific address that will not change between updates? If not, can we put it at a fixed address?

Currently not yet, but it can be place in specific region following the cf2 (basically the std way of uf2 for configures) https://github.com/microsoft/uf2/blob/master/cf2.md . However, don't worry too much about that for now. The bootloader can perform self-update by drag&drop `update--imxrt1010-evk.uf2` to the msc disk. The update.uf2 is an application that have the whole bootloader binary as array and will erase and flash the qspi with new one, then perform self-destruct to prevent flashing all over again. Note: I know this may not work well with protected bootloader that forbid unprotect, so yes we will need other way to update it later on. However, as long as It could update itself. We are relatively future-proof for now. 

> This image can be loaded into flash or RAM and can be loaded without additional debugger hardware. If we implement the automated revision check (or blank check), it should be faster to load through USB than through SWD.

Yes, thank you very much for clarification. I am definitely need to update my knowledge on this. I just found this app note here https://www.nxp.com/docs/en/nxp/application-notes/AN12238.pdf . I will do follow up review and let you know the update.

nxf58843:
> Could you please tell me which format/command that we could use the pycod with to flash the bootloader to qspiflash.

The [load command](https://github.com/pyocd/pyOCD/blob/master/docs/command_reference.md#load) includes an address for binary files.

> That is great to hear, I just want to confirm that after the bootloader is flashed, and bootmode = normal. The uf2 bootloader will run without the need of sdhost to load it to SRAM right ?

Yes, after the image is in flash, in normal boot mode, the boot ROM will check the headers and copy the image into RAM and the UF2 bootloader will start automatically.

The self update process is a little simpler when running from RAM.  You can generate a UF2 file directly from the binary image and drag-n-drop it onto the UF2 boot drive.  It is already running from RAM so it can load applications or bootloader updates directly to flash.  There is no need to build a special update image, just convert the binary to UF2.

hathach:
> The [load command](https://github.com/pyocd/pyOCD/blob/master/docs/command_reference.md#load) includes an address for binary files.

Ah right, I will update the flash-pyocd target to use the binary format instead. 

> Yes, after the image is in flash, in normal boot mode, the boot ROM will check the headers and copy the image into RAM and the UF2 bootloader will start automatically.

Superb !! I have done more testing and everything work perfect. I have known a bit more about the bootROM after reading the flash loader appnote

image

 
> The self update process is a little simpler when running from RAM. You can generate a UF2 file directly from the binary image and drag-n-drop it onto the UF2 boot drive. It is already running from RAM so it can load applications or bootloader updates directly to flash. There is no need to build a special update image, just convert the binary to UF2.

I agree, having the whole bootloader in SRAM allow us to directly flash the bootloader region. the above update-uf2 is actually a common scheme for all ports, since not all MCU has such large SRAM and comprehensive bootROM like rt10xx

hathach (approved):
Thank you very much for this awesome and comprehensive PR. This is indeed complicated feature that required person with lots of time working with the MCU to pull it off. Therefore it took me a bit of time to understand of the whole picture !!!!

There is a few places where I think we could  refactor the code like update flash-pyocd, generate bootloader.uf2, merge load2ram doc to the main port rt10xx readme.md as how-to for the port (every ports has its special way) and fix ci etc .... However, those are minor and need a bit of time for testing, I will probably do a series of follow-up myself. Thanks again for this great PR.

PR #48, opened 2020-12-28
Feature lpc55 updates

nxf58843:
I added more functionality to the LPC55 implementation.
This builds and I can get the debug logs on the LPCXpresso55S69 board.
USB connects (LED turns green when plugged in), but the MSC drive is not recognized by the PC.
It will only connect to the full speed USB port (USB0).  I tried modifying tusb_config.h but it never connects with the HS port.


ports/lpc55/board_flash.c
```diff
@@ -23,31 +23,94 @@
...
+static uint32_t _flash_page_addr = NO_CACHE;
+static uint8_t  _flash_cache[FLASH_PAGE_SIZE] __attribute__((aligned(4)));
+
 //--------------------------------------------------------------------+
 //
 //--------------------------------------------------------------------+
 void board_flash_init(void)
 {
-
+	if (FLASH_Init(&_flash_config) == kStatus_Success) {
```
hathach, line 47: It is minor, if possible please keep the tab size = 2 spaces.

ports/lpc55/tusb_config.h
```diff
@@ -39,6 +39,8 @@
 #endif
 
 #define CFG_TUSB_RHPORT0_MODE      OPT_MODE_DEVICE
+//#define CFG_TUSB_RHPORT0_MODE      OPT_MODE_NONE
+//#define CFG_TUSB_RHPORT1_MODE      (OPT_MODE_DEVICE | OPT_MODE_HIGH_SPEED)
```
hathach, line 43: please remove these temp code

ports/lpc55/boards.h
```diff
@@ -31,6 +31,9 @@
 
 #include "board.h"
 
+// Flash Start Address of Application
+#define BOARD_FLASH_APP_START  0x10000
```
hathach, line 35: 64KB is quite a lot for this mcu, what is our current output size ? If possible, 32KB is more ideal.

ports/lpc55/boards.c
```diff
@@ -95,7 +85,7 @@ void board_dfu_init(void)
   RESET_PeripheralReset(kUSB1_RST_SHIFT_RSTn);
   RESET_PeripheralReset(kUSB1RAM_RST_SHIFT_RSTn);
 
-#if (defined USB_DEVICE_CONFIG_LPCIP3511HS) && (USB_DEVICE_CONFIG_LPCIP3511HS)
+#if 1 || (defined USB_DEVICE_CONFIG_LPCIP3511HS) && (USB_DEVICE_CONFIG_LPCIP3511HS)
```
hathach, line 88: please remove temp code `1 ||`

ports/lpc55/boards.c
```diff
@@ -123,34 +113,44 @@ void board_dfu_complete(void)
   NVIC_SystemReset();
 }
 
+/* Defined in board_flash.c
```
hathach, line 116: would you mind telling why this function can't be implemented here ?

ports/lpc55/boards/lpcxpresso55s69/board.c
```diff
@@ -0,0 +1,116 @@
...
+#define IOCON_PIO_FUNC7          0x07u   /*!<@brief Selects pin function 7 */
+#define IOCON_PIO_INV_DI         0x00u   /*!<@brief Input function is not inverted */
+#define IOCON_PIO_MODE_INACT     0x00u   /*!<@brief No addition pin function */
+#define IOCON_PIO_MODE_PULLUP    0x20u   /*!<@brief Selects pull-up function */
+#define IOCON_PIO_OPENDRAIN_DI   0x00u   /*!<@brief Open drain is disabled */
+#define IOCON_PIO_SLEW_STANDARD  0x00u   /*!<@brief Standard mode, output slew rate control is enabled */
+
+#define IOCON_PIO_DIG_FUNC0_EN   0x0100u /*!<@brief Digital pin function 0 enabled */
+#define IOCON_PIO_DIG_FUNC1_EN   0x0101u /*!<@brief Digital pin function 1 enabled */
+#define IOCON_PIO_DIG_FUNC2_EN   0x0102u /*!<@brief Digital pin function 2 enabled */
```
hathach, line 53: These values are fixed for lpc55, unless they are changed from variant to variant, we should put them in the family `lpc55/boards.h`

hathach (changes requested):
Thank you very much for the PR, there are only a few minor feedback as below. Please also make sure
- The PR can have FS port enumerated OK as it is in the master
- Please fix the ci build error for lpc55 69 and 28 https://github.com/adafruit/tinyuf2/runs/1618730540?check_suite_focus=true . You don't have to implemented the 55s28, only need to put stub to pass the ci build.

> USB connects (LED turns green when plugged in), but the MSC drive is not recognized by the PC.
Hmm, I got the board enumerate and show up as disk drive just fine on Linux. Can you confirm the current code in the master without this pr modification ?

> It will only connect to the full speed USB port (USB0). I tried modifying tusb_config.h but it never connects with the HS port.
Don't worry about the HS port for now, the tinyusb stack probably need some minor changes to get it work. I haven't got time to do that, since there are no interest on lpc55 from anyone until now.

nxf58843:
Have you committed and pushed all your edits?  I get build errors for lpc55 on master.

```
usb_descriptors.c:195:38: error: 'serial_id[_3]' may be used uninitialized in this function [-Werror=maybe-uninitialized]
  195 |           uint8_t nibble = (serial_id[i] >> (j * 4)) & 0xf;
      |                                      ^
```

if I fix that ( `uint8_t serial_id[16] = {0};` ), I still get this error:

```
LINK _build/lpcxpresso55s69/tinyuf2-lpcxpresso55s69.elf
ld: /tmp/tinyuf2-lpcxpresso55s69.elf.tHfXUq.ltrans0.ltrans.o: in function `CLOCK_SetPLL1Freq':
/tinyuf2/lib/nxp/sdk/devices/LPC55S69/drivers/fsl_clock.c:1765: undefined reference to `SDK_DelayAtLeastUs'
ld: /tmp/tinyuf2-lpcxpresso55s69.elf.tHfXUq.ltrans0.ltrans.o: in function `CLOCK_EnableUsbfs0DeviceClock':
/tinyuf2/lib/nxp/sdk/devices/LPC55S69/drivers/fsl_clock.c:1901: undefined reference to `SDK_DelayAtLeastUs'
```

ports/lpc55/boards.c
nxf58843: Reading erased flash causes a bus fault due to ecc error.  You need to use FLASH_Read() to check.  I moved this to board_flash.c so that all the fsl_iap routines are together.  That is why I put board_usb_get_serial  there as well since FFR_GetUUID is in fsl_iap_ffr.h  I could add the headers to boards.c if that makes more sense.

ports/lpc55/boards.h
nxf58843: With all the debug logs turned on it is already about 32kB.  We can probably get to 32KB, but I would like to keep some headroom for development.

ports/lpc55/boards/lpcxpresso55s69/board.c
nxf58843: This was legacy from an example.  I will move these.

ports/lpc55/board_flash.c
nxf58843: I will update to be more consistent.

hathach:
what is your armgcc version ? the master ci build just fine with  lpc 55s69, so I guess  I pushed everything up already 
https://github.com/adafruit/tinyuf2/actions/runs/443695457

```
$ arm-none-eabi-gcc --version
arm-none-eabi-gcc (xPack GNU Arm Embedded GCC, 64-bit) 9.3.1 20200408 (release)
Copyright (C) 2019 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

Also make sure you have all the submodules update as well.

ports/lpc55/boards.h
hathach: the default size should be around 32KB, you can temporarily increases the size with DEUBG LOG build.

ports/lpc55/boards.c
hathach: thank you, that make sense, no problem at all. We can refactor this later on, I just wonder why would you move it.

hathach:
I am able to reproduce the issue with `serial_id` build error, gcc message is a bit descriptive. I am trying to figure out the reason.

ports/lpc55/boards.h
hathach: Maybe trying -Os would help as well, though I think at this state, we could just leave it with 64KB for easy debug. Though I would hope to have it around 32KB for release and working version.

hathach:
update I found the issue, gcc is super smart, it is able to detect the array overflow, but somehow can only do this with lpc55 port while other also return 16 bytes of serial length and still sneak through. Probably M33 has more aggressive optimization

@nxf58843 would you mind changing this line https://github.com/adafruit/tinyuf2/blob/master/src/usb_descriptors.c#L164 to

```
static uint16_t _desc_str[32+1];
```

that `+1` will allow us this PR to pass the CI.

PS: let's me know if you got MSC disk appear after the change. If yes then I think we could merge this PR for now, other functionality can be added via another PR.

nxf58843:
I updated usb_descriptors.c to fix the CI.
It builds and is connecting to USB.  I see the MSC drive appear in linux, but windows is not showing the drive for some reason.

hathach:
> I updated usb_descriptors.c to fix the CI.
> It builds and is connecting to USB. I see the MSC drive appear in linux, but windows is not showing the drive for some reason.

The CI still failed, you probably overwrite some basic board_ API with the forced push. I am sure it will be a quick fix. Regarding the windows issues, I actually didn't test with windows previously. I will do a test later on after the new year break. Will let you know the result of current master, if the issues exist previously, then we can just merge this. I am actually very familiar with this type of issue between host OS, so it isn't too hard to fix.

nxf58843:
Finally fixed the CI.  It was the self-update build that was failing.
Double tap is working and the drive shows up in linux.

hathach:
> Finally fixed the CI. It was the self-update build that was failing.
> Double tap is working and the drive shows up in linux.

Great works, can you also update the 55s28 board so that It could pass the ci build as well. You don't have to implement it, just add stub/missing function so that it could build.

nxf58843:
LPC55S28 is building and should be functional now.

src/main.c
```diff
@@ -115,9 +118,12 @@ static bool check_dfu_mode(void)
 int main(void)
 {
   board_init();
-
   TU_LOG1("TinyUF2\r\n");
 
+  // Flash needs to be initialized to check for a valid image
+  TU_LOG2("Flash init...\r\n");
+  board_flash_init();
```
hathach, line 125: [MANDATORY] you cannot change the order of function in main.c , this will affect other ports. If it is needed, added init function that must be called in either board_init() or board_dfu_init()

ports/lpc55/boards.c
```diff
@@ -23,9 +23,12 @@
  */
 
 #include "board_api.h"
+#include "boards.h"
+#include "board.h"
```
hathach, line 27: should only include `boards.h` only, no need for the specific `board.h`

ports/lpc55/boards/lpcxpresso55s28/board.h
```diff
@@ -54,22 +59,26 @@
 #define USB_VID           0x1fc9
 #define USB_PID           0x0094
 #define USB_MANUFACTURER  "NXP"
-#define USB_PRODUCT       "LPCXpresso 55s28"
+#define USB_PRODUCT       "LPCXpresso 55s69"
```
hathach, line 62: just simple typo for s28

src/usb_descriptors.c
```diff
@@ -180,7 +180,7 @@ uint16_t const* tud_descriptor_string_cb(uint8_t index, uint16_t langid)
 
     case STRID_SERIAL:
     {
-      uint8_t serial_id[16];
+      uint8_t serial_id[16] = {0};
```
hathach, line 183: we can revert this change now since it is not really needed

hathach (changes requested):
Thank you for lots of work on this port. There is only a request to revert the order of initialization in main.c since this will affect other ports. Other are just minor and optional, which I could do it myself later on as follow up.

src/main.c
nxf58843: This is critical for this port, and may be needed in other ports as well.  You cannot check the validity of an app in flash without initializing the flash interface.  This order is better for the iMXRT10xx port as well.  It is less critical for most devices with internal flash, except for LPC55 which has bus error when reading erased flash.

hathach: if `FLASH_Init` is critical within lpc55 that must be called, you could just place it in the `board_init()` for now. `board_flash_init()` is mostly for preparing flash for writing op, which is why I currently put it after `board_dfu_init()`

nxf58843: Does it cause any harm to call flash init first?
If you are running from RAM, you would need to initialize the flash before jumping to the application.  This is a difference between pushing code into RAM with sdphost vs letting the boot ROM copy the code into RAM on i.MXRT.  The boot ROM pre configures the flash for reading, but you need to do this manually if you load with sdphost.   If they load the image with sdphost, they will most often be updating the bootloader, but it seems safer to initialize the flash so the behavior is the same and we do not prevent them from jumping to an image in flash.  

I can make this change if you think there will be adverse side effects for other platforms.

nxf58843: What if instead of calling board_flash_init in main, we call it from board_init, or board_dfu_init depending on the platform requirements?  
This would require more changes, but would be cleaner than implementing with #ifdef

hathach: @nxf58843 that is possible as well but not in the scope of this PR. If you are willing to make the changes and test it on all other platforms, feel free to make a separate PR for the changes.

nxf58843: I moved boar_flash_init into board init for the mimxrt10xx and lpc55 platforms, and added `#ifndef EARLY_FLASH_INIT` for other platforms in main.c to get this PR through.  This can be simplified in a future PR.

ports/mimxrt10xx/boards.c
```diff
@@ -101,6 +101,9 @@ void board_init(void)
 
   LPUART_Init(UART_DEV, &uart_config, freq);
 #endif
+
+  // Flash needs to be initialized to check for a valid image
+  board_flash_init();
```
hathach, line 106: imxrt doesn't need this to call `board_flash_init()` to get it work

src/main.c
```diff
@@ -134,6 +130,13 @@ int main(void)
 
   TU_LOG2("DFU init...\r\n");
   board_dfu_init();
+
+// Some platforms initialize flash earlier to check for a valid image
+#ifndef EARLY_FLASH_INIT
```
hathach, line 135: we should completely remove the board_flash_init() and leave it up to `board_init()` or `board_dfu_init()` to does its work.

src/main.c
hathach: thank you for the changes, though I don't like the introduction of `EARLY_FLASH_INIT`, we should completely remove the `board_flash_init()` making it the internal port function that either is called within board_init() or board_dfu_init() if needed.

hathach: PS: I probably didn't make myself clear enough, sorry for that. I mean we should completely remove the board_flash_init() by making it part of either board_init() (early) or board_dfu_init() (late) decided by each port. You don't have to it as part of this PR. just get it passed while keeping the current order of execution by having extern internal_flash_init() to call `FLASH_Init()` and implement `board_flash_init()` as empty. I will do a cross-port PR to refactor it as follow-up later on.

ports/mimxrt10xx/boards.c
nxf58843: It is safer to initialize the flexspi in board_init.  Flexspi is not initialized when using serial downloader mode, so doing this here will make the behavior consistent in serial downloader mode and internal boot mode.

src/main.c
nxf58843: Ok, technically speaking, initializing flash earlier for i.MXRT10xx would make the firmware behave more consistently if it gets copied to RAM in internal boot mode by the ROM, or if it is pushed in through serial downloader boot mode, since the serial downloader boot mode would not initialize the FlexSPI interface.  But, in this case, behaving differently is a happy accident.  Not initializing the flash before checking the app validity means that it will never find a valid app when you push the image in through serial downloader mode, but that is probably the safest way to ensure that you can recover the bootloader image.  If you are pushing an image in through serial downloader mode, the odds are that you want to update the bootloader, and double tap reset will not work because it will go back to serial downloader mode.  Thus, to be able to update when you have a valid app and a broken bootloader, the safest thing to do is to always enter dfu mode whenever the image was loaded through the serial downloader, and initializing the flash after the validity check accomplishes that.  I'll work on cleaning this up.

hathach: wow, this is indeed a happy accident, look like we are very lucky for this one :)

hathach (approved):
Superb !! Thank you very much for putting lots of effort into lpc55 port. This has been a really great collaboration experience. I am looking forward to your upcoming PRs as well.

PR #49, opened 2020-12-29
rt10xx link2ram followup

hathach:
follow up to #41 (still WIP) 
- rt10xx use bootloader bin for self-update
- add flash pyocd with bin file target
- should fix action ci for self-update. Note: failed to build with rt1020 since it is not ported to link2ram just yet.
- add convenient `flash-sdp` along with sdphost binaries
- clean up, refactor linker symbols
- fix ci build thanks to #50 

@nxf58843 ready for review, if you have time to take a quick look. Regarding the note

> 2. Implement bootloader version, so when the image is loaded through SDP, it will self copy to flash if there is no flashloader or an older flashloader image in flash.  It simply needs to copy itself to flash.

I think we don't really need to have it to check the version and update, we should perform a simple `board_bootlaoder_valid()` check similar to https://github.com/adafruit/tinyuf2/blob/master/ports/mimxrt10xx/boards.c#L193 . If it is not present (fcb, ivt, interrupt make no sense), we could just simply copy bootloader from ram to flash.

The update when newer take more effort but not really needed, since once the bootlaoder is present, people generally won't use sdphost anymore. And can simply download the `update-tinyuf2.uf2` from latest release and drag and drop that to update the bootloader. SDPHost will be mostly for initial flash and recovery, I guess.


ports/mimxrt10xx/Makefile
```diff
@@ -58,22 +52,59 @@ INC += \
...
+			SDPHOST = sdphost/linux/amd64/sdphost
+	  endif
+
+	  ifeq ($(UNAME_I),i386)
+			SDPHOST = sdphost/linux/i386/sdphost
+	  endif
+	endif
+endif
+
+load-sdp: $(BUILD)/$(OUTNAME).bin
```
hathach, line 96: @nxf58843 this will make life easier for us, though I can only find the sdphost for x86, **do you know how or where to get the sdphost for ARM/linux e.g RPi**. Adafruit preferably use RPi instead of PC for production since it is cheaper, consume less power and basically easier to set up.

nxf58843: Compiled versions for linux, mac and windows are included in the tools/sdphost folder of the [IMX-RT1050-FLASHLOADER](https://www.nxp.com/webapp/Download?colCode=IMX-RT1050-FLASHLOADER&appType=license)

To run on the RPi, you would need to build from the source.  The source code is included in the SDK if you include the "MCU Boot" middleware option under SDK_2.8.2_EVK-MIMXRT1010/middleware/mcu-boot/tools/tools/sdphost/sdphost/

Here is a link to the [RT1010 SDK for linux with all middleware](https://mcuxpresso.nxp.com/download/88af2ba86fae2830192edc464d095333)

nxf58843:
@hathach 
> I think we don't really need to have it to check the version and update, we should perform a simple `board_bootlaoder_valid()` check similar to https://github.com/adafruit/tinyuf2/blob/master/ports/mimxrt10xx/boards.c#L193 . If it is not present (fcb, ivt, interrupt make no sense), we could just simply copy bootloader from ram to flash.
> 
> The update when newer take more effort but not really needed, since once the bootlaoder is present, people generally won't use sdphost anymore. And can simply download the `update-tinyuf2.uf2` from latest release and drag and drop that to update the bootloader. SDPHost will be mostly for initial flash and recovery, I guess.

I was just thinking the same thing.  I'll add a check for the FCFB and copy the image to flash if it is not present.

ports/mimxrt10xx/Makefile
hathach: superb !! Thank you very much, I will include the src since it is also BSD License. ( I am too lazy to do crosscompile for RPi for now).

hathach:
Thank you @nxf58843 very much for helping me out tremendously with this follow up, I am happy with the current code state and think it is good enough for the merge and ready for next release of tinyuf2

PR #50, opened 2020-12-29
added load2ram support for RT1020

nxf58843:
I added load2ram support for the RT1020 consistent with the RT1060 and RT1010.
This builds, but I do not have hardware to test.


hathach (approved):
superb!! Thank you very much, I have tried it on my 1020-evk and it works perfectly

PR #51, opened 2020-12-30
auto-copy when FCFB not present

nxf58843:
Rt10xx followup pr41 to automate copying image from RAM to flash if flash does not contain an FCFB block as determined by checking for FLEXSPI_CFG_BLK_TAG at FlexSPI_AMBA_BASE + 0x400U


hathach (approved):
superb !! Thank you very much, this will be very helpful. I think we need to set indicator for visual effect when flashing. That is easy, I will do it after the merge.

PR #54, opened 2021-01-09
generate uf2 artifact for feather

ladyada:


hathach:
I always forgot to update the ci yml :sweat_smile:

PR #55, opened 2021-01-10
STM32F303 Port

ReFil:
This is a port for the stm32f303 discovery board, but will be expended in the future, I Started with the f4 code and rewrote the HAL functions to use the F3 functions as opposed to the older F4 functions. Unfortunately it doesn't compile, there is some linker error complaining of problems with the tinyusb libraries despite it working perfectly for other boards. I think it's the final hurdle to make it work
image


ports/stm32f3/Makefile
```diff
@@ -0,0 +1,79 @@
...
+	$(ST_HAL_DRIVER)/Src/stm32f3xx_hal_cortex.c \
+	$(ST_HAL_DRIVER)/Src/stm32f3xx_hal_rcc.c \
+	$(ST_HAL_DRIVER)/Src/stm32f3xx_hal_gpio.c \
+	$(ST_HAL_DRIVER)/Src/stm32f3xx_hal_flash.c \
+	$(ST_HAL_DRIVER)/Src/stm32f3xx_hal_flash_ex.c \
+	$(ST_HAL_DRIVER)/Src/stm32f3xx_hal_uart.c
+
+SRC_C += \
+	$(PORT_SRC_C) \
+	lib/tinyusb/src/portable/st/synopsys/dcd_synopsys.c \
```
hathach, line 44: stm32f303 use different dcd than f4 change this to `lib/tinyusb/src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c`

ports/stm32f3/Makefile
```diff
@@ -0,0 +1,79 @@
...
+CFLAGS += \
+  -flto \
+  -mthumb \
+  -mabi=aapcs \
+  -mcpu=cortex-m4 \
+  -mfloat-abi=hard \
+  -mfpu=fpv4-sp-d16 \
+  -nostdlib -nostartfiles \
+  -DCFG_TUSB_MCU=OPT_MCU_STM32F3 \
+	-Wno-uninitialized
```
hathach, line 23: please move the SDK warning suppress to line 26

ports/stm32f3/Makefile
```diff
@@ -0,0 +1,79 @@
+UF2_FAMILY_ID = 0x57755a57
```
hathach, line 1: F3 family ID is `0x6b846188`

hathach (commented):
Superb, this is already a great start for F3 port. I notice you haven't specified the correct TinyUSB DCD driver. It is probably cause for compiling issue. Also could you 
- rename the board from `stm32f303vc` to `stm32f303disco` if discovery is the board you are working on.
- Add it to  ci matrix build here for build check https://github.com/adafruit/tinyuf2/blob/master/.github/workflows/build.yml#L113

Let me know if you need further help with the port.

ReFil:
Got it building now, there were a few other issues that needed fixing too, my USB clock initialisation was throwing errors and didn't look important so i got rid of it, a few defines i missed in tusb_config.h and there was an error related to _board_dfu_dbl_tap which i realised needed defining in the linker script. i'm about to flash it to my discovery board now and see if it works

ports/stm32f3/boards.h
```diff
@@ -0,0 +1,58 @@
...
+#include "stm32f3xx.h"
+#include "stm32f3xx_hal_conf.h"
+
+#include "board.h"
+
+// Flash Start Address of Application
+#define BOARD_FLASH_APP_START   0x08010000
+
+// Double Reset tap to enter DFU
+#define USE_DFU_DOUBLE_TAP      1
```
hathach, line 41: change this `USE_DFU_DOUBLE_TAP to 0` to disable the double tap feature so that you could get it compiled. First step should be usb enumerated.

hathach:
That is a great start, note I just push a bunch of minor macro rename. It is advisable that once you got it enumerated, do a pull from master to get it sync up.

ReFil:
It's not being detected as a usb device. The led is lit though so the code is doing something, do you have any ideas? Many thanks

hathach:
you probably need to check the clock, pin, setup, usb irq forward, macro config etc ... best is to compare it to currently running board from my stack repo here https://github.com/hathach/tinyusb/tree/master/hw/bsp/stm32f303disco

ReFil:
Looking at that I see a new differences. The usb clock initialisation is something I tried but it wouldn't compile when I added it so I removed it

ReFil:
image
this is the issue i get when i include the USB clock configuration stuff, It's really confusing me as the hall stuff is all corect

ReFil:
Ok I figured out the usb clock problem, I needed to include rcx_ex. It now initialises as a usb composite device for a second then disconnects and shows up as "device descriptor request failed"

ReFil:
https://user-images.githubusercontent.com/31960031/104233584-07cb2900-544a-11eb-82d6-d5ddaf69ee96.mp4

here's a video of what it does. Also the github builds keep failing despite my builds working properly which is very unusual

hathach:
I would suggest to try to enable UART peripheral and build with `LOG=2` (more in readme.md) to see if there is any issues. You could also try to run cdc_msc from tinyusb repo to verify your hardware + os combo is fine https://github.com/hathach/tinyusb/tree/master/examples/device/cdc_msc

ReFil:
Ok thank you, I already tried one of the mass storage examples and it worked perfectly so the issue is my code

ReFil:
Ok there's some voodoo magic going on. It now shows up as a mass storage device but I have to push reset twice slowly. I still don't know why the GitHub build actions aren't working but I'll sort that next/clean up the code. Then I'm going to check the actual flashing works

hathach:
> Ok there's some voodoo magic going on. It now shows up as a mass storage device but I have to push reset twice slowly. I still don't know why the GitHub build actions aren't working but I'll sort that next/clean up the code. Then I'm going to check the actual flashing works

it failed due to the `self-update` target, you have to move those interrupt handler inside the `#ifndef TINYUF2_SELF_UPDATE
`

ports/stm32f3/boards.c
```diff
@@ -267,25 +267,7 @@ void OTG_FS_IRQHandler(void)
 
 #endif
 
-void USB_HP_IRQHandler(void)
```
hathach: these IRQ should be remained, `OTG_FS_IRQHandler()` is the one should be removed.

hathach (commented):
`self-update` target don't use USB at all, you need to move those IRQ inside the `#ifndef TINYUF2_SELF_UPDATE` . Please also correct your tab/indent size to 2 spaces

ReFil:
looking at this i think board_flash needs to be changed as the flash sector mapping seems totally different to the f4 
image
is the size argument of the struct sector size in bits?

ReFil:
yeah it's totally different, the f3 memory is only 2k wide and has 128 pages, whereas the f4 has 11 pages of between 16 and 128kbit
image
https://www.st.com/resource/en/reference_manual/dm00031020-stm32f405-415-stm32f407-417-stm32f427-437-and-stm32f429-439-advanced-arm-based-32-bit-mcus-stmicroelectronics.pdf
image

ReFil:
ok it's not flashing, i'm attempting to flash the stm32 discovery example hex converted using uf2conv but it's not working, one issue i believe is the start location as it's trying to start at 0x08000000 which is where the bootloader resides, it should be installing at 0x08010000. it's also possible that my board flash code is just utterly buggered. I still cant get debug data out, I've tried the st link built into the discovery with a jumper from pb3 to pin 6 of the st link header and a dedicated USB uart at 115200 and i'm not getting anything out

ReFil:
It's now erasing the flash space correctly and programming the flash properly, however it's not entering the app

ReFil:
I've tried a new enterting routine and now it wont enter the bootloader at all, double tap doesn't work either. Is there some trick to getting debug messages as even when building with logger = swo and log=2 it wont output debug messages

hathach:
> I've tried a new enterting routine and now it wont enter the bootloader at all, double tap doesn't work either. Is there some trick to getting debug messages as even when building with logger = swo and log=2 it wont output debug messages

You need to make sure the UART is enabled and configured properly. The LOG > 1 will output message default to UART, which is probably connected to stlink on discovery board.

ReFil:
I'm so close with this, the last problem is it wont initialise the USB device properly on boot. i'm trying to boot into an application that also does USB device and somehow during board_init the board pops up as a USB device with dsescriptor request failed and that stopsw it from showing up once it enters the app too. the only fix is to keep it powered through the stlink USB as i reconnect the user USB at which point i can jump back and forth between the app and bootloader with no trouble, as well as flash new software and have it enter that software fine. I could only get it to go from app to bootloader and vice versa by calling tud_disconnect in the board jump to app and board init and as a result the github workflow doesn't work
also I've noticed i need to push the reset button 6 times before it shows up as a USB drive which strikes me as odd

ReFil:
ok It now properly initalises the usb device when it boots, and i can swap back and forth between the app and the bootloader however when i go from app to bootloader it doesn't enumerate

ReFil:
I can solve the app to bootloader enumeration problem however because i need to tap it 6 times to get it to show up as a USB drive this breaks everything. There is some issue with the USB enumeration and i have no idea what it is

ReFil:
I am officially out of ideas, if I completely shut off the board then power it up nothing works, I get device descriptor failure. I have to keep the board powered then remove and reinsert the usb before it shows up. Furthermore i still haven't found why I need to reboot the board 3 times before it shows up as a usb drive

ReFil:
What role does the vbus sense and id pins play? The discovery board I'm playing with doesn't have these

ReFil:
I soldered a bodge wire to d+ and can force a re enumeration successfully by shorting it to gnd but what I don't understand is why my code in board_app_jump doesn't do this itself

hathach:
HI, I am currently busy with other works, will try to check this out later on when having more time.

ReFil:
I would appreciate that as i am all out of ideas

hathach:
still busy with RP2040, get back to this later on. Thanks for your patient.

ReFil:
Still stuck with this, been trying a few hacks in the meantime but even when i forcibly re enumerate it when it's in the bootloader it doesn't show up as mass storage until i reset it 3 times

hathach:
@ReFil I have fixed the issues with usb enumeration due to incorrect UUID address for getting serial string descriptor. Only correct the UART port config for f3 discovery board along with other clean up. I haven't got time to test with actual uf2 flashing or other functionality. I will leave it to you, when you thing the PR is ready for merging, please convert it to Ready state. It would be great if you could also provide an application uf2 for testing as well at that time.

ReFil:
Thank you so much. I will endeavour to finish the job. I'll check and test everything and then provide a uf2 for testing when I believe it's complete

hathach:
No problems at all, I am glad that I could help.

ReFil:
Had to adjust the cache size to make the flashing work again and i failed at reading the reference manual and got the erase start point of the flash wrong so it wouldn't flash over a working firmware
[zmk.zip](https://github.com/adafruit/tinyuf2/files/5921731/zmk.zip)
. Now it boots into bootloader, flashes cleanly whether starting from an erased flash or a working program. Attached are two uf2s of ZMK, a software intended for keyboards. An easy way of determining functionality is hooking up a neopixel to spi2 (PB15), it'll turn red or whitish depending on which file you flash. it might also try and turn your volume up uncontrollably because the buttons arent setup right so be careful, don't wear headphones when testing!

ports/stm32f3/boards.c
```diff
@@ -0,0 +1,327 @@
...
+
+void board_app_jump(void)
+{
+  GPIO_InitTypeDef  GPIO_InitStruct;
+  GPIO_InitStruct.Pin = GPIO_PIN_12;
+  GPIO_InitStruct.Mode = GPIO_MODE_OUTPUT_PP;
+  GPIO_InitStruct.Pull = GPIO_NOPULL;
+  GPIO_InitStruct.Speed = GPIO_SPEED_FREQ_HIGH;
+  HAL_GPIO_Init(GPIOA, &GPIO_InitStruct);
+  HAL_GPIO_WritePin(GPIOA, GPIO_PIN_12, GPIO_PIN_RESET);
```
hathach, line 141: This shouldn't be needed, since app_jump is invoked before dfu_init() therefore usb isn't inited at this time.

hathach (commented):
Superb work !! I made some clean up to the PR myself. Have tested the uf2 flashing, it works well. The double tap works as well. I haven't looked at why increasing MSC buf cause board_flash failed to work. It normally just make copying uf2 file faster. But this is more than enough for initial PR for new port. There is one review I would like you to remove before merging though since I don't think it is needed.

ports/stm32f3/boards.c
ReFil: I've just tested it and it is required, my USB device refuses to initialise after bootloader without it, I suspect it's because without being actively pulled down it gets left floating and windows tries to enumerate it. As for the msc cache size i just copied the value from the f303 discovery msc sample code for tinyusb.

hathach: OK I guess we will leave it for now. There will be more time and PR to enhance f3 port

hathach (approved):
Excellent work, thank you very much for your effort and patient. This works well on both double tap and uf2 flashing which are the main feature of this repo.

PR #56, opened 2021-01-11
moved neopixel driver to submodule shared with tinyUSB and other proj…

gsteiert:
Moved neopixel driver for double m33 express to submodule shared with tinyUSB and other projects.  Other LPC55 boards will be able to make use of this and get updates to the common source.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -22,3 +22,6 @@
 [submodule "lib/uf2"]
 	path = lib/uf2
 	url = https://github.com/microsoft/uf2.git
+[submodule "lib/sct_neopixel"]
+	path = lib/sct_neopixel
+	url = https://github.com/gsteiert/sct_neopixel
--- a/ports/lpc55/boards/double_m33_express/board.c
+++ b/ports/lpc55/boards/double_m33_express/board.c
@@ -29,110 +29,11 @@
 #include "fsl_gpio.h"
 #include "fsl_power.h"
 #include "fsl_iocon.h"
-#include "fsl_sctimer.h"
+#include "sct_neopixel.h"
 
 #include "clock_config.h"
 
-//--------------------------------------------------------------------+
-// Neopixel Driver
-//--------------------------------------------------------------------+
-#define NEO_SCT           SCT0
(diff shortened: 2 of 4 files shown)


hathach (approved):
Great, thank you for the PR, that would help to implement the neopixel driver on other MCU. Just a quick question, can we use SCT on the iMXRT ? Also currently it is not the focus off the tinyuf2 just yet, but eventually we would need sctpix_teardown/deinit() API before jumping to app.

hathach:
you need to add the sct c file to the self update source as well to fix the ci

gsteiert:
> Great, thank you for the PR, that would help to implement the neopixel driver on other MCU. Just a quick question, can we use SCT on the iMXRT ? Also currently it is not the focus off the tinyuf2 just yet, but eventually we would need sctpix_teardown/deinit() API before jumping to app.

The iMXRT does not have SCTimer,  It has FlexIO, but we need to check if that is available on the NeoPixel pin.

I'll add de-init to the to-do list.  Will people have the option to keep features pre-initialized going into the app?  I know MakeCode has been looking at keeping the MSC active through the jump to the app so you could initiate an update without having to double tap.  Odds are most apps will want to enable the same features on the board.

hathach:
> The iMXRT does not have SCTimer, It has FlexIO, but we need to check if that is available on the NeoPixel pin.

no need to worry about that, current iMXRT use DWT for delay and GPIO for bitbang, and work really great.
 
> I'll add de-init to the to-do list. Will people have the option to keep features pre-initialized going into the app? I know MakeCode has been looking at keeping the MSC active through the jump to the app so you could initiate an update without having to double tap. Odds are most apps will want to enable the same features on the board.

I am not a fan of it, IMHO, bootloader should tear down everything it init() before jumping so that app could have a clean state to start with. There is no rush, other ports doesn't tear down all of its peripheral for now. I was focusing on the self-update first, since once we get it done, we can fix all existing/future bugs of bootloader with ease.

hathach (approved):
Thank you for your PR. This would help adding NXP MCU with SCT easier in the future

PR #58, opened 2021-01-11
Add feather esp32s2 tft

hathach:
Try to add feahter esp32s2 tft with LCD display standard bootloader image from makecode (can be customized later on). Following is the code running on esp32s2 Kaluga board. Since I don't have the board, @ladyada  would you mind doing a quick test if having time.

Note: the rotation, font size etc .. can be off. Though I could only hope it does display something !!

image


ladyada (approved):

ladyada:
let me try it now (note that long term i may change pins)

ladyada:
hmm it does not enumerate ....

hathach:
> hmm it does not enumerate ....

It is probably caused by untested tft code, the kaluga LCD is based on ILI9341. I have just disabled the TFT for now, will wait until the lcd breakout to deliver for further testing :).

hathach:
@ladyada I got it running with breakout 1.14" , which the same wiring as schematics (note the Backlight active state is HIGH, you may need to change it to 0 if needed). The existing font size for header seems to be too big for 1.1.4 (color format seems to be off as well) . I will try to change/adapt  it.  

image

ladyada:
hmm backlight does not come on, it should be set output/high to turn on

hathach:
just push the update for pin definition to `TFT rev` the `normal TFT` is probably out dated I guess.

hathach:
Offset is correct but color still be off, look like inverted but not quite right
 ![Uploading 104494963...]![Uploading 104494964...]

ladyada:
looks better! yes check for MADCTL register (color order)

hathach:
everything looks great now. @ladyada let me know if it works for you. I will do a bit of refactor, also central the firmware.uf2 --> FTHRS2BOOT Icon.  

image

PR #60, opened 2021-01-21
Initialize UART fully for logging.

petejohanson:
Will trying to port this to some f401 boards I have, I was having a hard time getting logging working. I finally tracked this down to the UART not being fully initialized on startup. This seemed to do it.

Still tracking down the remaining issues, but at least have logs now!


hathach (approved):
Ah right, my bad, I seem to forget enable the UART properly after moving code around :)

PR #76, opened 2021-02-15
Fixing FCFB address on RT1060

nxf58843:
This corrects the FCFB address on RT1060.  
The RT1060 expects the FCFB at 0x0.  The existing code places it at 0x400, but does not overwrite 0x0 - 0x400, so if an image was loaded there previously with a valid FCFB it would work anyway.
This also moves the double tap to LPGPR which works better with the POR button SW3 on the RT1060 EVK, but the timing for the POR button is sensitive because the delay after releasing the button is longer than SW9.


hathach (approved):
excellent works as usual. Thank you very much of fixing the issue with rt1060. Pulled out the board and tested, It works much better now, don't worry about the button timing. I found it perfectly find to  use double tap with the SW9.

PR #80, opened 2021-03-06
Fix build on rp2040

kaetemi:
Just the small changes needed to make it work. Using this to flash a peripheral.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -138,7 +138,7 @@ int main(void)
   screen_draw_drag();
 #endif
 
-#if CFG_TUSB_OS == OPT_OS_NONE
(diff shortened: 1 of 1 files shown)


hathach:
It is indeed a small change, though since rp2040 support uf2 in bootrom. I am not sure whether we will ever port it into this repo. Are you doing rp2040 ?

kaetemi:
We're using this to flash a peripheral that's connected to an rp2040. Not rp2040 itself. (Uploading our tool to the rp2040 in no_flash mode, flashing our peripheral, and then rebooting back into rp2040 bootrom.) Greatly simplifies workflow. :)

hathach (approved):
Thanks for the context, look good to me.

PR #81, opened 2021-03-06
Add support for multiple sectors per cluster in GhostFAT

kaetemi:
Add support for multiple sectors per cluster in GhostFAT to enable larger flash sizes.

### DIFF
--- a/src/ghostfat.c
+++ b/src/ghostfat.c
@@ -89,7 +89,7 @@ struct TextFile {
 //--------------------------------------------------------------------+
 
 #define BPB_SECTOR_SIZE           ( 512)
-#define BPB_SECTORS_PER_CLUSTER   (   1)
+#define BPB_SECTORS_PER_CLUSTER   (CFG_UF2_SECTORS_PER_CLUSTER)
 #define BPB_RESERVED_SECTORS      (   1)
 #define BPB_NUMBER_OF_FATS        (   2)
 #define BPB_ROOT_DIR_ENTRIES      (  64)
@@ -98,13 +98,12 @@ struct TextFile {
 #define FAT_ENTRY_SIZE            (2)
 #define FAT_ENTRIES_PER_SECTOR    (BPB_SECTOR_SIZE / FAT_ENTRY_SIZE)
 // NOTE: MS specification explicitly allows FAT to be larger than necessary
-#define BPB_SECTORS_PER_FAT       ( (BPB_TOTAL_SECTORS / FAT_ENTRIES_PER_SECTOR) + \
-                                   ((BPB_TOTAL_SECTORS % FAT_ENTRIES_PER_SECTOR) ? 1 : 0))
+#define TOTAL_CLUSTERS_ROUND_UP   ((BPB_TOTAL_SECTORS + BPB_SECTORS_PER_CLUSTER - 1) / BPB_SECTORS_PER_CLUSTER)
+#define BPB_SECTORS_PER_FAT       ((TOTAL_CLUSTERS_ROUND_UP + FAT_ENTRIES_PER_SECTOR - 1) / FAT_ENTRIES_PER_SECTOR)
 #define DIRENTRIES_PER_SECTOR     (BPB_SECTOR_SIZE/sizeof(DirEntry))
 #define ROOT_DIR_SECTOR_COUNT     (BPB_ROOT_DIR_ENTRIES/DIRENTRIES_PER_SECTOR)
 
 STATIC_ASSERT(BPB_SECTOR_SIZE                              ==       512); // GhostFAT does not support other sector sizes (currently)
-STATIC_ASSERT(BPB_SECTORS_PER_CLUSTER                      ==         1); // GhostFAT presumes one sector == one cluster (for simplicity)
 STATIC_ASSERT(BPB_NUMBER_OF_FATS                           ==         2); // FAT highest compatibility
 STATIC_ASSERT(sizeof(DirEntry)                             ==        32); // FAT requirement
 STATIC_ASSERT(BPB_SECTOR_SIZE % sizeof(DirEntry)           ==         0); // FAT requirement
@@ -167,7 +166,7 @@ STATIC_ASSERT( CLUSTER_COUNT >= 0x1015 && CLUSTER_COUNT < 0xFFD5 );
 #define UF2_SIZE           (UF2_SECTORS * BPB_SECTOR_SIZE)
 
 #define UF2_FIRST_SECTOR   ((NUM_FILES + 1) * BPB_SECTORS_PER_CLUSTER) // WARNING -- code presumes each non-UF2 file content fits in single sector
-#define UF2_LAST_SECTOR    ((UF2_FIRST_SECTOR + UF2_SECTORS - 1) * BPB_SECTORS_PER_CLUSTER)
+#define UF2_LAST_SECTOR    (UF2_FIRST_SECTOR + UF2_SECTORS - 1)
 
 #define FS_START_FAT0_SECTOR      BPB_RESERVED_SECTORS
 #define FS_START_FAT1_SECTOR      (FS_START_FAT0_SECTOR + BPB_SECTORS_PER_FAT)
@@ -270,11 +269,11 @@ void uf2_read_block (uint32_t block_no, uint8_t *data)
     for ( uint32_t i = 0; i < FAT_ENTRIES_PER_SECTOR; ++i )
     {
       // Generate the FAT chain for the firmware "file"
-      uint32_t v = (sectionIdx * FAT_ENTRIES_PER_SECTOR) + i;
+      uint32_t v = (sectionIdx * FAT_ENTRIES_PER_SECTOR * BPB_SECTORS_PER_CLUSTER) + (i * BPB_SECTORS_PER_CLUSTER);
 
       if ( UF2_FIRST_SECTOR <= v && v <= UF2_LAST_SECTOR )
       {
-        ((uint16_t*) (void*) data)[i] = v == UF2_LAST_SECTOR ? 0xffff : v + 1;
+        ((uint16_t*) (void*) data)[i] = v == UF2_LAST_SECTOR ? 0xffff : (v / BPB_SECTORS_PER_CLUSTER) + 1;
       }
     }
   }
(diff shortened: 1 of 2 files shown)


kaetemi:
Affects the size of the FAT table (each entry points to a cluster, not to a sector), so the FAT table is smaller with a larger sector per cluster value. Affects the generation of FAT table entries (clusters instead of sectors.)

Affects the size of the text files. Each file takes a cluster at minimum, not a sector. This affects the offset of the current.uf2 file.

Writing is not affected.

kaetemi:
For flashing Bridgetek EVE graphics controller (as peripheral).

It includes both program and graphics data. Support is up to 256MB flash.

kaetemi:
Yep, data between repeated read and write matches. Text and html file look good too.

hathach (approved):
Superb !! Thank you very much for the PR. It greatly increase the virtual disk capacity. I hope more user will find it useful since it is really neat but rarely see in mcu world.

kaetemi:
FAT16 is relatively simple.
* Header (the struct with all the size values)
* FAT (just a big array of uint16, one for each cluster - a cluster is just a combined bunch of physical sectors to reduce the size of this array (and speed up allocation) - the value of the uint16 is either empty, EOF, or pointing to the next cluster of the file)
* Root directory (just an array of file entries, one or more sectors, can be smaller than a cluster)
* Data (the clusters, may also contain directories, a file is minimum one cluster since that's the allocation granularity)

hathach:
Thank you for detail explanation, I also understand the basic but it is still complicated though 🤯

src/ghostfat.c
```diff
@@ -270,11 +269,11 @@ void uf2_read_block (uint32_t block_no, uint8_t *data)
     for ( uint32_t i = 0; i < FAT_ENTRIES_PER_SECTOR; ++i )
     {
       // Generate the FAT chain for the firmware "file"
-      uint32_t v = (sectionIdx * FAT_ENTRIES_PER_SECTOR) + i;
+      uint32_t v = (sectionIdx * FAT_ENTRIES_PER_SECTOR * BPB_SECTORS_PER_CLUSTER) + (i * BPB_SECTORS_PER_CLUSTER);
 
       if ( UF2_FIRST_SECTOR <= v && v <= UF2_LAST_SECTOR )
```
henrygab, line 274: Because `v` is the `fatArrayIndex`, this should use `UF2_FIRST_CLUSTER_NUMBER` and `UF2_LAST_CLUSTER_NUMBER`.  Those are not defined, but recommended in prior comment.

src/ghostfat.c
```diff
@@ -270,11 +269,11 @@ void uf2_read_block (uint32_t block_no, uint8_t *data)
...
       {
-        ((uint16_t*) (void*) data)[i] = v == UF2_LAST_SECTOR ? 0xffff : v + 1;
+        ((uint16_t*) (void*) data)[i] = v == UF2_LAST_SECTOR ? 0xffff : (v / BPB_SECTORS_PER_CLUSTER) + 1;
```
henrygab, line 276: Reverting this line is recommended.  Your changes were based on the very confusing `#define` names used in the prior code, coupled with your commit's change to the semantic meaning of two of the `#define`s above.

<details><summary>Comments to help refresh my memory</summary><P/>

Presuming the above change to use `UF2_FIRST_CLUSTER_NUMBER` and `UF2_LAST_CLUSTER_NUMBER`....

Here, `v` is the `fatArrayIndex` being modified by the particular loop iteration.  The array index inside the FAT uses the "cluster number" as its index.

The correct values to insert here are the `cluster number`.
This appears to be setting the FAT entry at index `v`


</details>

Revert this line, keeping the simpler:

```C
((uint16_t*) (void*) data)[i] = v == UF2_LAST_CLUSTER_NUMBER ? 0xffff : v + 1;
```

src/ghostfat.c
```diff
@@ -319,16 +318,16 @@ void uf2_read_block (uint32_t block_no, uint8_t *data)
   else if ( block_no < BPB_TOTAL_SECTORS )
   {
     sectionIdx -= FS_START_CLUSTERS_SECTOR;
-    if ( sectionIdx < NUM_FILES - 1 )
+    if ( sectionIdx < (NUM_FILES - 1) * BPB_SECTORS_PER_CLUSTER )
```
henrygab, line 321: <details><summary>Comments to help refresh my memory</summary><P/>

* `sectionIdx` is renamed to `sectionRelativeSector`
* The root directory in FAT16 is NOT part of the data volume
* Therefore, if only having files in root as ghostfat does, the first data sectors are for the file contents.

</details>

Yes, good!

Please update the static assertion to reflect that each file in ghostfat must fit within a single cluster, not a single sector (e.g., ~lines 142 here)

src/uf2.h
```diff
@@ -42,8 +42,12 @@ SOFTWARE.
 // Version is passed by makefile
 // #define UF2_VERSION         "0.0.0"
 
-#define CFG_UF2_FLASH_SIZE  (4*1024*1024)   // TODO absolute max flash size
-#define CFG_UF2_NUM_BLOCKS  0x10109         // just under 32MB
+#define CFG_UF2_FLASH_SIZE          (4*1024*1024)    // TODO absolute max flash size
+#define CFG_UF2_NUM_BLOCKS          (0x10109)        // just under 32MB
+#define CFG_UF2_SECTORS_PER_CLUSTER (1)
+// #define CFG_UF2_FLASH_SIZE          (256*1024*1024)  // TODO absolute max flash size
```
henrygab, line 48: Remove the commented-out defines.  Instead, if your intent is to define a highest value tested / supported, just make that a comment in the file.

henrygab (commented):
@kaetemi,  Excellent work.

@hathach, there are a number of changes I would recommend.  Some of them strongly.

Let me know if any questions.

src/ghostfat.c
```diff
@@ -98,13 +98,12 @@ struct TextFile {
 #define FAT_ENTRY_SIZE            (2)
 #define FAT_ENTRIES_PER_SECTOR    (BPB_SECTOR_SIZE / FAT_ENTRY_SIZE)
 // NOTE: MS specification explicitly allows FAT to be larger than necessary
-#define BPB_SECTORS_PER_FAT       ( (BPB_TOTAL_SECTORS / FAT_ENTRIES_PER_SECTOR) + \
-                                   ((BPB_TOTAL_SECTORS % FAT_ENTRIES_PER_SECTOR) ? 1 : 0))
+#define TOTAL_CLUSTERS_ROUND_UP   ((BPB_TOTAL_SECTORS + BPB_SECTORS_PER_CLUSTER - 1) / BPB_SECTORS_PER_CLUSTER)
```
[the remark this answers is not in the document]
kaetemi: Interesting, thanks.

src/ghostfat.c
```diff
@@ -167,7 +166,7 @@ STATIC_ASSERT( CLUSTER_COUNT >= 0x1015 && CLUSTER_COUNT < 0xFFD5 );
 #define UF2_SIZE           (UF2_SECTORS * BPB_SECTOR_SIZE)
 
 #define UF2_FIRST_SECTOR   ((NUM_FILES + 1) * BPB_SECTORS_PER_CLUSTER) // WARNING -- code presumes each non-UF2 file content fits in single sector
-#define UF2_LAST_SECTOR    ((UF2_FIRST_SECTOR + UF2_SECTORS - 1) * BPB_SECTORS_PER_CLUSTER)
+#define UF2_LAST_SECTOR    (UF2_FIRST_SECTOR + UF2_SECTORS - 1)
```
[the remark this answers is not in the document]
kaetemi: It's defined as *UF2_FIRST_SECTOR **+ UF2_SECTORS** - 1*.
The multiplication multiplies UF2_SECTORS (the number of sectors) by sectors per cluster. That does not make sense.

src/ghostfat.c
```diff
@@ -270,11 +269,11 @@ void uf2_read_block (uint32_t block_no, uint8_t *data)
     for ( uint32_t i = 0; i < FAT_ENTRIES_PER_SECTOR; ++i )
     {
       // Generate the FAT chain for the firmware "file"
-      uint32_t v = (sectionIdx * FAT_ENTRIES_PER_SECTOR) + i;
+      uint32_t v = (sectionIdx * FAT_ENTRIES_PER_SECTOR * BPB_SECTORS_PER_CLUSTER) + (i * BPB_SECTORS_PER_CLUSTER);
```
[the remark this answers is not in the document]
kaetemi: Math at 274 looks right. Sector `v` is aligned to the cluster boundary (so it's rounded down, essentially), which means that should always be less or equal to the actual last sector.

But yes, as you specified before, it'll be simpler working with the clusters rather than sectors there.

kaetemi:
@henrygab With the standardized terminology it'll be clearer, yes. Thanks for the suggestions. I'll do a cleanup tomorrow.

henrygab:
@kaetemi ... I listed the wrong PR... mine WIP PR is #84. Would appreciate if you could try out the changes from my PR.  The changes pass the automated builds. 

FYI, I think your changes had some unexpected results, including effectively duplicating the small files' data (first sector per cluster) on all the other sectors of that same cluster.  Not visible from the file system, but visible from a dump of the volume.

I think my PR addresses all the above items.  Let me know if you agree.  Unfortunately, I don't currently have my setup to build/test the changes "live" on a device.  In the past, I've added code to manually exercise the ghostfat code, and dump the results to a disk image.  I then used 010 editor (a hex editor with ability to parse structures) to look at FAT volumes, and found it helpful.

I lost that wrapper code.   Maybe, if @hathach agrees, I can create a proper stub that generates well-defined patterns of data via board_flash_read() and dumps the full volume... with the goal of allowing automated testing of this ghostfat code (by generating images and running custom comparison that excludes directory entry date/time stamp changes).

hathach:
> I lost that wrapper code. Maybe, if @hathach agrees, I can create a proper stub that generates well-defined patterns of data via board_flash_read() and dumps the full volume... with the goal of allowing automated testing of this ghostfat code (by generating images and running custom comparison that excludes directory entry date/time stamp changes).

Thank you @henrygab for taking a deep look at this PR, I am totally clueless. For the stub, sure, you could just put it somewhere in the ghostfat as well if you are not familiar with the file org of the repo. This is more like an generic version across MCU based on nRF52. We are trying to standardize and make bootloader project easier for users/makers.  

```C
#if 0
// stub test code
#endif 
``` 

Later on we could move it to its own tests if we manage to add unit test suite support to this repo as well.

PR #84, opened 2021-03-11
Update for PR81

henrygab:
Fixes a few bugs.
Renames some variables for clarity / to avoid confusion.


hathach:
@henrygab actually they allow it to switch back (right around the reviewers name), maybe you only need repo permission. I just sent an collaborator invitation, if you are interested in the project, please consider to accept it. 

image

henrygab:
Wow!  I certainly didn't notice that before, but you're absolutely right.  Thanks for giving me better options! :)

hathach:
no problems, we all learn something new from each other :)

PR #85, opened 2021-03-11
Update pr 81

henrygab:
Clarifies various sections of code.
Fixes a few bugs.
Enhanced long-term maintainability.

Most significantly, this enhances the consistent use of terms "sector" vs. "Cluster Number" vs. "Cluster Index".
* `Sector` refers to the sector, relative to the BPB.
    * If the media is partitioned, this is the sector offset within the partition.
    * For unpartitioned media, the sector is equivalent to the logical block address (LBA).
* `Cluster Number` is what FAT defines as the same term.  The first cluster number of the data section is .... **_TWO_** ... due to how FAT evolved over the decades.
* `Count of Clusters` is the count of actual data clusters in the data section.  In other words, if considering all the data section as an array of clusters, this is the count of full clusters usable.
* `Cluster Index` is not actually used; but I suggest it in case one needs to treat the clusters as an array, and wants a name for variables that index into that array (to avoid having to constantly subtract two from a cluster number when indexing the array).


src/ghostfat.c
```diff
@@ -269,23 +275,28 @@ void uf2_read_block (uint32_t block_no, uint8_t *data)
...
     {
       // Generate the FAT chain for the firmware "file"
-      uint32_t v = (sectionIdx * FAT_ENTRIES_PER_SECTOR * BPB_SECTORS_PER_CLUSTER) + (i * BPB_SECTORS_PER_CLUSTER);
+      uint32_t v = (sectionRelativeSector * FAT_ENTRIES_PER_SECTOR * BPB_SECTORS_PER_CLUSTER) + (i * BPB_SECTORS_PER_CLUSTER);
 
-      if ( UF2_FIRST_SECTOR <= v && v <= UF2_LAST_SECTOR )
+      if ( UF2_FIRST_CLUSTER_NUMBER <= v && v < UF2_LAST_CLUSTER_NUMBER )
       {
-        ((uint16_t*) (void*) data)[i] = v == UF2_LAST_SECTOR ? 0xffff : (v / BPB_SECTORS_PER_CLUSTER) + 1;
+        ((uint16_t*) (void*) data)[i] = v + 1; // contiguous file, so point to next cluster number
```
hathach, line 282: slip up the last sector make it much easier to read, though the value of `v` doesn't seem to be the same meaning. Should we drop the multiply with `BPB_SECTORS_PER_CLUSTER` since it is cluster num and not sector num

src/ghostfat.c
```diff
@@ -312,29 +331,44 @@ void uf2_read_block (uint32_t block_no, uint8_t *data)
...
-      memcpy(data, info[sectionIdx / BPB_SECTORS_PER_CLUSTER].content, strlen(info[sectionIdx / BPB_SECTORS_PER_CLUSTER].content));
+      // WARNING -- code presumes first data cluster == first file, second data cluster == second file, etc.
+      struct TextFile const * inf = &info[ sectionRelativeSector / BPB_SECTORS_PER_CLUSTER ];
+      size_t fileContentStartOffset = (sectionRelativeSector % BPB_SECTORS_PER_CLUSTER) * BPB_SECTOR_SIZE;
+      size_t fileContentLength = strlen(inf->content);
+
+      if (fileContentLength > fileContentStartOffset) {
+        size_t bytesToCopy = fileContentLength - fileContentStartOffset;
+        const char * dataStart = (inf->content) + fileContentStartOffset;
+        STATIC_ASSERT(sizeof(char) == 1); // technically not required to be true
```
hathach, line 352: this should always be true for embedded, though i don't really needs it. We could just declare dataStart as either `const void*` or `const uint8_t*`

hathach (commented):
Thanks for the great PR as usual. The rename make everything clearer. I have tried it on esp32s2, however, seem like the `v` value at line 282 is mis-computed at bit. Here is what I got in the dmesg

```
[337632.000452] sd 6:0:0:0: [sda] Attached SCSI removable disk
[337632.553736] FAT-fs (sda): error, fat_bmap_cluster: request beyond EOF (i_pos 8275)
[337632.553741] FAT-fs (sda): Filesystem has been set read-only
[337632.553748] FAT-fs (sda): error, fat_bmap_cluster: request beyond EOF (i_pos 8275)
```

kaetemi:
I'll try it on hardware here.

kaetemi:
Seems to work with the following patches applied.

```patch
@@ -171,8 +171,8 @@ STATIC_ASSERT( CLUSTER_COUNT >= 0x1015 && CLUSTER_COUNT < 0xFFD5 );
                             ((UF2_SECTOR_COUNT % BPB_SECTORS_PER_CLUSTER) ? 1 : 0))
 #define UF2_BYTE_COUNT     (UF2_SECTOR_COUNT * BPB_SECTOR_SIZE) // always a multiple of sector size, per UF2 spec
 
-#define UF2_FIRST_CLUSTER_NUMBER ((NUM_FILES + 1 + 2))
-#define UF2_LAST_CLUSTER_NUMBER  (UF2_FIRST_CLUSTER_NUMBER + UF2_CLUSTER_COUNT - 1)
+#define UF2_FIRST_CLUSTER_INDEX ((NUM_FILES + 1))
+#define UF2_LAST_CLUSTER_INDEX  (UF2_FIRST_CLUSTER_INDEX + UF2_CLUSTER_COUNT - 1)
 
 #define FS_START_FAT0_SECTOR      BPB_RESERVED_SECTORS
 #define FS_START_FAT1_SECTOR      (FS_START_FAT0_SECTOR + BPB_SECTORS_PER_FAT)
```

```patch
@@ -280,13 +280,13 @@ void uf2_read_block (uint32_t block_no, uint8_t *data)
     for ( uint32_t i = 0; i < FAT_ENTRIES_PER_SECTOR; ++i )
     {
       // Generate the FAT chain for the firmware "file"
-      uint32_t v = (sectionRelativeSector * FAT_ENTRIES_PER_SECTOR * BPB_SECTORS_PER_CLUSTER) + (i * BPB_SECTORS_PER_CLUSTER);
+      uint32_t v = (sectionRelativeSector * FAT_ENTRIES_PER_SECTOR) + i;
 
-      if ( UF2_FIRST_CLUSTER_NUMBER <= v && v < UF2_LAST_CLUSTER_NUMBER )
+      if ( UF2_FIRST_CLUSTER_INDEX <= v && v < UF2_LAST_CLUSTER_INDEX )
       {
         ((uint16_t*) (void*) data)[i] = v + 1; // contiguous file, so point to next cluster number
       }
-      else if ( v == UF2_LAST_CLUSTER_NUMBER)
+      else if ( v == UF2_LAST_CLUSTER_INDEX )
       {
         ((uint16_t*) (void*) data)[i] = 0xffff; // end of file marker in FAT16
       }

```

src/ghostfat.c
henrygab: <details><summary>Good question.... Let me think this through a bit.</summary><P/>

`sectionRelativeSector` here is the sector offset within the FAT table.
`i` here is the sector-specific FAT entry in the table (treating it as an array for just this single sector).

Therefore, `v` should simply be `sectionRelativeSector * FAT_ENTRIES_PER_SECTOR`, so it reflects the index into the FAT (across all sectors).
</details>

I think the right fix is to modify the line that declares `v` as follows:
```C
uint32_t v = (sectionRelativeSector * FAT_ENTRIES_PER_SECTOR)
```

The rest of the code seems to remain.

src/ghostfat.c
henrygab: OK, I will update to one of these, likely `const void *` unless you have other preference.

henrygab:
@kaetemi , Wow!  Thank you for finding a workaround and code that appears to work.

I am confused by the use of the UF2_FIRST_CLUSTER_INDEX macros causing this to work here.  I am therefore not entirely comfortable with this first part of the patch (yet).

The first for setting `uint32_t v = (sectionRelativeSector * FAT_ENTRIES_PER_SECTOR) + i;` is 100% spot-on.  This was my error, and both yourself and @hathach noted the correct fix for this line.

I think I will have to rewrite the wrapper to generate a disk image (with generated data corresponding to the sector being read), and then parse the results manually with 010 editor, unless I can otherwise understand why this feels wrong, and be comfortable with the use of the `..._INDEX` values instead.

My concern is that the file is pointing into the FAT two clusters later than it should.  I just don't have confidence that this should be using `..._INDEX` macros inside the FAT generating code:
* FAT[0] and FAT[1] are special, but otherwise...
* FAT[`n`] should store the next CLUSTER_NUMBER in the FAT chain, after CLUSTER_NUMBER == `n`.
* Thus, until the end of a contiguously allocated file, the FAT chain should store `FAT[n]  == n+1`
* Here, the array index `i` is sector-relative, so `v` is created as the absolute index into the fat array.

Could you dump a full disk image of the exposed ghostfat drive, and place it somewhere for download?  It'd be great if both an 
8 sector / cluster and a 1 sectors / cluster image were available, but that's actually optional.

I'll continue to look into this more.

src/ghostfat.c
hathach: I would use `const void*` as well.

hathach:
take your time, there is no rush at all :)

kaetemi:
Yeah, perhaps it's off by 2 elsewhere. I'll prepare some disk dumps.

kaetemi:
Start of the FAT table.

```
F8 FF FF FF FF FF FF FF 05 00 06 00 07 00 08 00 09 00 0A 00 0B 00 0C
```

kaetemi:
Oh.

```
((NUM_FILES + 1))
```

NUM_FILES is 3, it includes the UF2 file. Then the +1 already makes this the cluster number.

So patch should just be...

```patch
                             ((UF2_SECTOR_COUNT % BPB_SECTORS_PER_CLUSTER) ? 1 : 0))
 #define UF2_BYTE_COUNT     (UF2_SECTOR_COUNT * BPB_SECTOR_SIZE) // always a multiple of sector size, per UF2 spec
 
-#define UF2_FIRST_CLUSTER_NUMBER ((NUM_FILES + 1 + 2))
+#define UF2_FIRST_CLUSTER_NUMBER ((NUM_FILES + 1))
 #define UF2_LAST_CLUSTER_NUMBER  (UF2_FIRST_CLUSTER_NUMBER + UF2_CLUSTER_COUNT - 1)
 
 #define FS_START_FAT0_SECTOR      BPB_RESERVED_SECTORS
 #define FS_START_FAT1_SECTOR      (FS_START_FAT0_SECTOR + BPB_SECTORS_PER_FAT)
```

```patch
     for ( uint32_t i = 0; i < FAT_ENTRIES_PER_SECTOR; ++i )
     {
       // Generate the FAT chain for the firmware "file"
-      uint32_t v = (sectionRelativeSector * FAT_ENTRIES_PER_SECTOR * BPB_SECTORS_PER_CLUSTER) + (i * BPB_SECTORS_PER_CLUSTER);
+      uint32_t v = (sectionRelativeSector * FAT_ENTRIES_PER_SECTOR) + i;
 
       if ( UF2_FIRST_CLUSTER_NUMBER <= v && v < UF2_LAST_CLUSTER_NUMBER )
       {
         ((uint16_t*) (void*) data)[i] = v + 1; // contiguous file, so point to next cluster number
       }
       else if ( v == UF2_LAST_CLUSTER_NUMBER)
       {
         ((uint16_t*) (void*) data)[i] = 0xffff; // end of file marker in FAT16
       }
```

kaetemi:
Repeated FAT table looks good too.

Text files are repeating themselves through the whole cluster, but it's not really an issue.

UF2 file looks good.

henrygab:
@kaetemi -- well done, well done indeed!  Yes, the issue was that `NUM_FILES` included the UF2 file, and thus the calculation of `v` was off ... by exactly two.  It was simply coincidence that it matched FAT's cluster number offset.

> Text files are repeating themselves through the whole cluster, but it's not really an issue.

Really?  I thought I already included a fix for that.  If any file was more than 512 bytes, it would cause ghostfat to crash when reading that file's sector, so the prior code was not robust.    Did your source include the following?

<details><summary>Expand for code</summary><P/>

```C
  else if ( block_no < BPB_TOTAL_SECTORS )
  {
    sectionRelativeSector -= FS_START_CLUSTERS_SECTOR;
    uint32_t sectionRelativeClusterIndex = sectionRelativeSector / BPB_SECTORS_PER_CLUSTER;

    if ( sectionRelativeClusterIndex < (NUM_FILES - 1) )
    {
      // WARNING -- code presumes first data cluster == first file, second data cluster == second file, etc.
      struct TextFile const * inf = &info[ sectionRelativeSector / BPB_SECTORS_PER_CLUSTER ];
      size_t fileContentStartOffset = (sectionRelativeSector % BPB_SECTORS_PER_CLUSTER) * BPB_SECTOR_SIZE;
      size_t fileContentLength = strlen(inf->content);

      // nothing to copy if already past the end of the file (only when >1 sector per cluster)
      if (fileContentLength > fileContentStartOffset) {
        // obviously, 2nd and later sectors should not copy data from the start
        const uint8_t * dataStart = (inf->content) + fileContentStartOffset;
        // limit number of bytes of data to be copied to remaining valid bytes
        size_t bytesToCopy = fileContentLength - fileContentStartOffset;
        // and further limit that to a single sector at a time
        if (bytesToCopy > BPB_SECTOR_SIZE) {
          bytesToCopy = BPB_SECTOR_SIZE;
        }
        memcpy(data, dataStart, bytesToCopy);
      }
    }
    // ... else generate UF2 file data on-the-fly
```

If you did have the above code, I would be concerned, because I tried to have the file contents not get duplicated across each sector of the cluster.

kaetemi:
I'll double check if that part got merged in.

henrygab:
@kaetemi / @hathach -- I've marked this PR as ready for review.  I must rely on your testing for now (I don't have any testing setup currently), but based on early results, it seems now fully functional.

TODO: Would be good to have a few "fake" boards that exercise multi-cluster and some additional options (e.g., a non-UF2 file that is > 512 bytes in size).

kaetemi:
Tested again. Text files are OK. Everything else looks good too.

Dumps attached.

[dump_large_part.zip](https://github.com/adafruit/tinyuf2/files/6128968/dump_large_part.zip)

[dump_small_part.zip](https://github.com/adafruit/tinyuf2/files/6128976/dump_small_part.zip)

hathach (approved):
Excellent work with detailed comments as usual. Thank you very much @henrygab and @kaetemi for further improving the ghostfat. I have tested with my esp32s2 board ( since i am working with it). Able to write uf2 using CURRENT.UF2 and it works perfectly. This will help lots of other.

PS: Let's me know if you want  to do more tests and/or add more commit, else I am happy to merge

hathach:
> TODO: Would be good to have a few "fake" boards that exercise multi-cluster and some additional options (e.g., a non-UF2 file that is > 512 bytes in size).

To be honest, I am not sure what is best way to make fake boards/ports, for non-UF2 may be @kaetemi can help with that by filling the info file with bunch of texts as long as its contents is less than 1 cluster :)

henrygab:
@hathach -- I have confirmed from @kaetemi 's dump files the following:

1. FAT generated for both appears accurate.
2. Large clusters (32k) no longer causes duplicated data in entire cluster.

@hathach -- I think this is ready to be merged.  However, as I submitted this PR, I will wait for you (or another contributor) to approve and merge.

hathach:
merged now since it got approved already, thanks again :+1:

PR #86, opened 2021-03-12
Speedup esp32 flash

hathach:
- Esp32s2 increase flash cache from 4K to 64KB. This allow esp to use block erase instead of sector erase and gain 2x
speed flashing
- Also skip erase & write if contents already matches.

```
Before PR

written 1249536 bytes in 19.38 seconds.
Speed : 64.48 KB/s

After PR

written 1249536 bytes in 9.69 seconds.
Speed : 128.95 KB/s
```


ladyada:
thanks, this is great!

PR #90, opened 2021-03-13
Allow  >1 sector per cluster

henrygab:
Enable configuration of the ghostfat options via compiler defines.


hathach (approved):
Thanks, there is still lots of TODO, we will move it to per ports/boards later on

PR #91, opened 2021-03-13
Allow overriding the compilation date/time

henrygab:
This is to simplify the process of generating repeatable ghostfat images from tests.

Supporting this will help automate validation that the resulting image files are unchanged.

### DIFF
--- a/src/compile_date.h
+++ b/src/compile_date.h
@@ -37,15 +37,13 @@
   #define REPRODUCIBLE_TIME __TIME__
 #endif
 
-
-
-#define __YEAR_INT__ ((( \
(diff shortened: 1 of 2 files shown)


hathach (commented):
the name is too long how about changing it to `FAT_DATE` and `FAT_TIME`, there is no need for underscore since these aren't system macro. In case of conflict we could add `TUF2_` prefix.

henrygab:
All macros are now have the prefix `COMPILE_`, and short names.

hathach (approved):
thank you, it looks perfect now.

PR #92, opened 2021-03-16
Correct detecting the right sdphost binary for linux

jepler:
On my Linux system, "uname -i" just prints "unknown". "uname -m" prints "x86_64".

### DIFF
--- a/ports/mimxrt10xx/Makefile
+++ b/ports/mimxrt10xx/Makefile
@@ -95,9 +95,8 @@ else
 	endif
 endif
 
-SDPHOST ?= false
-
 flash-sdp: $(BUILD)/$(OUTNAME).bin
+	@if [ -z "$(SDPHOST)" ]; then echo SDPHOST is not found for this machine; exit 1; fi
 	$(SDPHOST) -u 0x1fc9,$(SDP_$(MCU)_PID) -- write-file $(SDP_$(MCU)_WRITE_ADDR) $<
 	$(SDPHOST) -u 0x1fc9,$(SDP_$(MCU)_PID) -- jump-address $(SDP_$(MCU)_JUMP_ADDR)
 
--- a/ports/mimxrt10xx/Makefile
+++ b/ports/mimxrt10xx/Makefile
@@ -79,22 +79,23 @@ ifeq ($(OS),Windows_NT)
 	SDPHOST = sdphost/win/sdphost.exe
 else
 	UNAME_S := $(shell uname -s)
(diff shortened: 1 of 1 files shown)


ladyada (approved):
i think its ok but will also wait for thach to reply!

ports/mimxrt10xx/Makefile
```diff
@@ -79,23 +79,24 @@ ifeq ($(OS),Windows_NT)
...
-	  ifeq ($(UNAME_I),i386)
+		endif
+		ifeq ($(UNAME_M),i386)
 			SDPHOST = sdphost/linux/i386/sdphost
-	  endif
+		endif
 	endif
 endif
 
+SDPHOST ?= false
```
hathach, line 98: maybe we should have a better error message with

```
ifndef SDPHOST
$(error SDPHOST is not found for this machine !!)
endif
```

hathach (commented):
It is odd, on my machine, both -i and -m has the same value `x86_64`. Accept for ?= false. I think it is good to merge

ports/mimxrt10xx/Makefile
jepler: Thanks, I improved this.  Now, if it's not detected,
```
$ make flash-sdp BOARD=imxrt1010_evk
SDPHOST is not found for this machine
```
however, other targets can still be built even if this happens.

hathach: that is even better

hathach (approved):
perfect now

jepler:
I researched the "why" a bit and found:

> 
> 
> POSIX doesn't define -p or -i. In GNU coreutils they are marked as non-portable, as you indicate; the default implementation relies on two optional operating system features, the three-argument form of sysinfo(2) (from SunOS) and the six-argument form of sysctl(3) (from the BSDs; neither of which are available on Linux).
> 
> Thus on Debian and derived distributions (apart from Ubuntu and its derivatives), you simply get unknown.
> 
> On Fedora and related distributions, uname is patched to return the machine type (-m) as processor (-p) and hardware platform (-i), with the latter tweaked to produce i386 for any value of the form i?86.
> 
> On Ubuntu and derivatives, a variant of the Fedora patch is used, which additionally checks for AMD CPUs on i686 processors and produces athlon instead.
> 

https://unix.stackexchange.com/questions/307955/uname-p-i-are-unknown

hathach:
Thanks @jepler it is good reading :+1: :+1:

PR #98, opened 2021-03-19
Rt1010 sdphost always write to flash

hathach:
Use bootmode from SRC->SBMR2 to determine whether we are in Serial Download mode. If so write the tinyuf2 image to flash. So basically everytime SDPHost is uesd, it will write contents to flash. This works better as 
- recovery method or when reworking components swapping mcu and/or flash chip. 
- prototyping/simulate other boards e.g using evk board without actually drag & drop uf2 once loaded via SDP
- At worse, user load SDP with the same image on flash, the write caching will compare with existing contents and won't perform any erase/write op.

@nxf58843 @gsteiert would you mind reviewing the PR, also please accept the contributor invitation, it make requesting you as reviewer much easier in the future

### DIFF
--- a/ports/mimxrt10xx/README.md
+++ b/ports/mimxrt10xx/README.md
@@ -18,17 +18,16 @@ make BOARD=metro_m7_1011 flash-jlink-bin
 
 iMXRT has built-in BootROM that implements the Serial Download Protocol (SDP), which can be used to load & execute TinyUF2 to SRAM with `spdhost` tool via USB. You need to
 
-1. Power up your board with the Boot Mode switch set to `BOOT_MODE[1:0]=01` to enter Serial Download mode. Note Serial Download mode also automatically run with blank flash, therefore you don't have to manual change it in your production run.
-2. Run `flash-sdp` make target which in turn uses the `sdphost` with correct address and arguments to load TinyUF2 
+1. Power up your board with the Boot Mode switch set to `BOOT_MODE[1:0]=01` to enter Serial Download mode. Note: Serial Download mode also automatically run with blank flash, therefore you don't have to manual change it in your production run.
+2. Run `flash-sdp` make target which in turn uses the `sdphost` with correct address and arguments to load and execute TinyUF2. While running, TinyUF2 will program the external flash with its SRAM's image.
 
   ```
   make BOARD=imxrt1010_evk flash-sdp
   ```
-3. (Optional) Drag and drop `update-tinyuf2_BOARD.uf2` to program the external flash. This step is optional for blank chip since TinyUF2 will automatically check the external flash to see if it has a valid Flash Configure Block (FCFB), if not TinyUF2 will program the external flash with its SRAM's image.
 
-Note: `sdphost` executable binaries for common platforms (windows/mac/linux) are included in sdphost folder. If you have issue with executable permission, just manually give it permission to run. The source code for sdphost tool is also included just in case binaries for your host platform is not included (e.g ARM, RISC-V etc ...)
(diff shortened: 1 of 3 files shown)


ports/mimxrt10xx/README.md
```diff
@@ -18,17 +18,16 @@ make BOARD=metro_m7_1011 flash-jlink-bin
...
 
-Note: `sdphost` executable binaries for common platforms (windows/mac/linux) are included in sdphost folder. If you have issue with executable permission, just manually give it permission to run. The source code for sdphost tool is also included just in case binaries for your host platform is not included (e.g ARM, RISC-V etc ...)
+Note1: `sdphost` executable binaries for common platforms (windows/mac/linux) are included in sdphost folder. If you have issue with executable permission, just manually give it permission to run. The source code for sdphost tool is also included just in case binaries for your host platform is not included (e.g ARM, RISC-V etc ...)
 
-Note2: Since SDP with BootROM doesn't requires external debugger and always exists regardless of the external flash, this method can also be used to de-brick your board should it needs. 
+Note2: Since SDP with BootROM doesn't requires external debugger and always exists regardless of the external flash, this method can also be used to de-brick your board should it needs.
```
nxf58843, line 30: "should it be necessary" or "should it be needed" would be better grammatically

ports/mimxrt10xx/board_flash.c
```diff
@@ -55,13 +55,15 @@ void board_flash_init(void)
 {
   flexspi_nor_flash_init(FLEXSPI_INSTANCE, (flexspi_nor_config_t*) &qspiflash_config);
 
-  // Check for FCFB and copy bootloader to flash if not present
-  if ( *(uint32_t*) FCFB_START_ADDRESS != FLEXSPI_CFG_BLK_TAG )
+  // TinyUF2 will copy its image to flash if  Boot Mode is '01' i.e Serial Download Mode (BootRom)
+  // Normally it is done once by SDPHost or used to recover an corrupted boards
+  uint32_t const boot_mode = (SRC->SBMR2 & SRC_SBMR2_BMOD_MASK) >> SRC_SBMR2_BMOD_SHIFT;
+  if (boot_mode == 1)
```
nxf58843, line 61: Should we check the revision to prevent accidentally downgrading?

nxf58843 (commented):
I will try to check on hardware this weekend.

ports/mimxrt10xx/board_flash.c
hathach: I thought of this, however there is actually no need to check against SDP. SDP is either production mode or the desperate mode. When used it should just overwrite whatever on the flash.

ports/mimxrt10xx/README.md
hathach: will fix it.

hathach:
@nxf58843 thank you for reviewing, I have tried it locally and it works well. Since there are several folks at Adafruit working on and need this to test out the prototype hardware. I will merge it now, should there is any issues, we will fix it with follow up PR.

ports/mimxrt10xx/board_flash.c
nxf58843: That is safest for de-bricking.  There is an outside chance that they grab an older .bin than was previously loaded on the board but if they switched to serial download mode it is probably safer to let them load it anyway.

hathach: yup, for version stepping, it is doable in much easier way with drag & drop `update-tinyuf2.uf2` is really the de-brick move when user accidentally corrupt flash config and/or bootloader

PR #100, opened 2021-03-19
metro m7: Switch bootloader to 4MB flash part

jepler:
& change product to match CircuitPython.  @ladyada I'm not sure whether it's this or a change in CircuitPython that is needed to fix copying CIRCUITPY content more than 1MB..


hathach (approved):
look good to me

ladyada (approved):

PR #101, opened 2021-03-19
add flash nuke-BOARD for iMX RT port

hathach:
adding flash nuke to build system

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -155,24 +155,24 @@ jobs:
 
     - name: Find Port
       run: |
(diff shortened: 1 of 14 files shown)


ladyada:
this worked on a metro m7!

hathach:
> this worked on a metro m7!

thanks for testing, CI failed miserably. Will wrap it up later on.

hathach:
merged this for now, after giving it more thought, I think we should have an erase-app which is more useful and much less dangerous than an erase-all (chip erase). What is useful of blank flash. Will do another PR for all port erase-firmware target much like the update- one.

PR #102, opened 2021-03-22
Add Imxrt erase firmware

### LINKED ISSUE #95: mimxrt: Add `nuke.bin` (and .uf2?)
@ladyada would like a "nuke" program for this platform.  It would just issue the general flash erase command, so that uf2 would have to be re-loaded via sdphost or j-link.

hathach:
This PR fix #95  
- ground work for `erase_firmware-BOARD.uf2` for all ports
- new board API: board_reset() and board_flash_erase_app()
- add erase_firmware as std application that reset board with MAGIC ERASE_APP
- IMXRT implement board_flash_erase_app() add makefile to support erase_firmware. For iMXRT it is much quicker to perform chip erase and re-write bootloader from sram to flash. (sector/block erase probably takes longer).
- `erase_firmware-BOARD.uf2` is also uploaded as ci artifact/release if available.
- Also update tinyusb to latest as well (why not).

### DIFF
--- a/ports/mimxrt10xx/Makefile
+++ b/ports/mimxrt10xx/Makefile
@@ -9,6 +9,8 @@ FLASH_BIN_ADDR = $(UF2_$(MCU)_WRITE_ADDR)
 
 include ../make.mk
 
+# TODO include port.mk
+
 MCU_DIR = lib/nxp/sdk/devices/$(MCU)
 LD_FILES ?= $(PORT_DIR)/linker/$(MCU)_ram.ld $(PORT_DIR)/linker/memory.ld $(PORT_DIR)/linker/common.ld
 
--- a/ports/mimxrt10xx/Makefile
+++ b/ports/mimxrt10xx/Makefile
@@ -66,6 +66,7 @@ SDP_MIMXRT1011_WRITE_ADDR = 0x20206400
 SDP_MIMXRT1011_JUMP_ADDR  = 0x20207000
 UF2_MIMXRT1011_WRITE_ADDR = 0x60000400
 
+# RT1015, 1050 pid is also 0x0130
 SDP_MIMXRT1021_PID = 0x0130
 SDP_MIMXRT1021_WRITE_ADDR = 0x0400
 SDP_MIMXRT1021_JUMP_ADDR  = 0x1000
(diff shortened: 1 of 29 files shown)


ports/mimxrt10xx/Makefile
```diff
@@ -97,7 +103,14 @@ else
...
 flash-sdp: $(BUILD)/$(OUTNAME).bin
-	@if [ -z "$(SDPHOST)" ]; then echo SDPHOST is not found for this machine; exit 1; fi
+	@:$(call check_defined, SDPHOST, not found for this machine)
+	$(SDPHOST) -u 0x1fc9,$(SDP_$(MCU)_PID) -- write-file $(SDP_$(MCU)_WRITE_ADDR) $<
+	$(SDPHOST) -u 0x1fc9,$(SDP_$(MCU)_PID) -- jump-address $(SDP_$(MCU)_JUMP_ADDR)
+
+# TODO write-register doesn't work
+erase-app: $(BUILD)/$(OUTNAME).bin
+	@:$(call check_defined, SDPHOST, not found for this machine)
+	$(SDPHOST) -u 0x1fc9,$(SDP_$(MCU)_PID) -- write-register $(DBL_TAP_REG_ADDR) 32 $(DBL_TAP_MAGIC_ERASE_APP)
```
hathach, line 113: @nxf58843 somehow I couldn't get the write-register to work, it always get no response from device. Though this is not important, most user should use the erase-firmware.uf2 . I just wonder if I made any mistakes. Else we could just remove this together. 

```
sdphost -u 0x1fc9,0x0145 -- write-register 0x400D410C 32 0xf5e80ab4
getStatusResponse.readPacket error 5.
Status (HAB mode) = 10004 (0x2714) No response from device.
```

read-register doesn't work as well

```
sdphost -u 0x1fc9,0x0145 -- read-register 0x400D410C
getStatusResponse.readPacket error 5.
Status (HAB mode) = 10004 (0x2714) No response from device.
```

nxf58843: Let me look into this.
Have you tried reading an address from OCRAM?
I need to see if that address is accessible or needs to be enabled.

hathach: I have just tried to read from dtcm and ocram but it doesn't response either. No problems at all, I am just curious, using uf2 is more preferable. I will merge it now, you don't have to "fix" it at all :smiley: 

```
$ sdphost -u 0x1fc9,0x0145 -- read-register 0x20200000
getStatusResponse.readPacket error 5.
Status (HAB mode) = 10004 (0x2714) No response from device.

$ sdphost -u 0x1fc9,0x0145 -- read-register 0x20000000
getStatusResponse.readPacket error 5.
Status (HAB mode) = 10004 (0x2714) No response from device.
```

nxf58843: Glad you can move forward with the uf2 implementation.
It looks like the read/write register commands were dropped from the RT1010 due to ROM size limitation.
You could write a single word using write-file, but adding another file is not much better than separate binaries.
I may try setting the magic numbers with different jump addresses when I can find time.

hathach: ah thanks for confirmation, I will just remove the make target. uf2 will be the standard way for erase_firmware (factory reset) for tinyuf2. Don't worry about having other register for erase_firmware. I don't see any issue with packing to the double tap reg. May only need to rename these variable a bit to be more generic than just double tap

PR #103, opened 2021-03-23
add sdphost binaries for rpi4 (tested)

hathach:
- Remove src since it is not buildable
- Add note to external source repo for building binary

@jepler added binaries for RPI4 (arm 32 bit), and tested on actual hardware as well, work well. The current src is indeed not compilible and removed. I give the link to https://github.com/apexrtos/nxp_blhost_sdphost instead.


jepler:
Thank you!

PR #94, opened 2021-03-17
Ghostfat test boards

henrygab:
The goal of this branch is to enable, as part of the normal build process, validation of GhostFat image generation.

* [x] Create framework to execute own custom `test_main()` function, using existing code
* [x] Ensure ghostfat initialization completes prior to calling `test_main()`
* [x] Generate disk image file from `test_main()`
* [x] Validate bit-for-bit accuracy of disk image file from `test_main()`
* [x] Use common framework code for variations of desired file system image
* [x] Check in known-good file system image(s) to validate against
* [x] Double-compress known-good disk images
* [x] Ensure CI catches failures

NOTE: When the ghostfat image that is generated *would* need to change, such as for a bugfix, then updated versions of the known-good file system images would also need to also be committed.

PR #104, opened 2021-03-24
Followup ghostfat test

hathach:
follow up to #94 to separate native build tests from cross-compile (for target hardware). `NO_TINYUF2_BUILD` is used for this purpose (along with SKIP_NANOLIB, some later ports especially non-ARM may skip this lib as well). NO_TINYUF2_BUILD introduced in recent PR to allow make.mk and rule.mk to build other executable image such as `erase_firmware` (will also migrate self_update later on). 

Eventually, test_ghostfat/main.c run on its own and interact with a small subset of board_api, those that are used by ghostfat.c. 

PS: I really like this and want to move this to its own unit tests folder, maybe later. Got to catch up with iMXRT board release  https://www.adafruit.com/product/4950


henrygab:
OK, will review....

ports/test_ghostfat/main.c
```diff
@@ -369,10 +369,12 @@ int DumpDiskImage(void) {
     return ERR_NONE;
 }
 
-int test_main(void)
+int main(void)
 {
     int r;
 
+    uf2_init();
```
henrygab, line 376: Very nice!  So much cleaner! :)

.github/workflows/build.yml
```diff
@@ -27,13 +27,8 @@ jobs:
     - name: Setup Python
       uses: actions/setup-python@v2
 
-    - name: Setup Node.js
-      uses: actions/setup-node@v1
-
     - name: Checkout code
       uses: actions/checkout@v2
-      with:
-        submodules: 'true'
```
henrygab, line 36: Oh, submodules are not required for this?  I will defer to you, just want to double-check.

henrygab (approved):
Well done!  Just two minor questions.  I can approve now, as they are so minor.

.github/workflows/build.yml
hathach: yeah, sub-modules basically is the tinyusb and other mcu driver, the native doesn't uses any of those. ci passed is the proof for that. Those mcu is actually very large for some such as nxp and esp-idf (both > 100MB)

ports/test_ghostfat/main.c
hathach: thanks, yup, I think we could move it to its own tests folder later on :)

ports/test_ghostfat/main.c
```diff
@@ -79,7 +79,7 @@ void DumpBuffer(uint64_t startingOffset, const void * sectorBuffer, uint16_t byt
             for (int k = 0; k < 16; k++) {
                 if (0 == k % 8) {
                     printf(" ");
-                }
+                }//------------- Board  -------------//
```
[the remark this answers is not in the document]
hathach: oops, I must accidentally paste text with middle mouse. Happen all the time

hathach:
thank you for reviewing

PR #107, opened 2021-03-26
Added board files for Artisense reference design RD00

m-byte:
VID/PID as assigned by Espressif (see https://github.com/espressif/usb-pids)


ports/esp32s2/boards/artisense_rd00/board.h
```diff
@@ -0,0 +1,75 @@
...
+// USB UF2
+//--------------------------------------------------------------------+
+
+#define USB_VID           0x303A
+#define USB_PID           0x80B0
+#define USB_MANUFACTURER  "Artisense"
+#define USB_PRODUCT       "Reference Design RD00"
+
+#define UF2_PRODUCT_NAME  USB_MANUFACTURER " " USB_PRODUCT
+#define UF2_BOARD_ID      "AS-RD00"
```
hathach, line 71: `Board-ID` should follow the format of uf2 https://github.com/microsoft/uf2#files-exposed-by-bootloaders

```
```

hathach (changes requested):
all look good, except the board id must follow the sdd format from uf2 specs.

ports/esp32s2/boards/artisense_rd00/board.h
m-byte: OK. Should be fixed now

hathach (approved):
look good now, thanks for your pr

m-byte:
Thanks

PR #109, opened 2021-04-12
mimxrt10xx: Also self-install when FCFB not valid

jepler:
When starting with a blank flash chip, the board will enter sdp bootloader mode even if the jumpers are in the FlexSPI boot position.  However, since b28b83a24990624961b1af80be8257a91e24f14e, tinyuf2 would not write itself to flash in this situation. Make it do so in both the "boot switch" or "fcfb invalid" cases.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -97,7 +97,7 @@ jobs:
       id: idf-cache
       with:
         path: ${{ github.workspace }}/.idf_tools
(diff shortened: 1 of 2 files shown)


hathach (commented):
Look good, though could you merge these 2 into 1 if. We can skip the log. It is not really needed.

```C
if ( (boot_mode == 1) ||  !fcfb_valid) 
{
    write_tinyuf2_to_flash();  
}
```

hathach (approved):
perfect, thank you for the ci cache fix as well.

PR #110, opened 2021-04-19
Add board definition for the ATMegaZero ESP32-S2

eddieespinal:
Added board definition for the ATMegaZero ESP32-S2

VID/PID as assigned by Espressif 
[https://github.com/espressif/usb-pids/blob/main/allocated-pids.txt](https://github.com/espressif/usb-pids/blob/main/allocated-pids.txt)


ladyada (approved):

ports/esp32s2/boards/atmegazero_esp32s2/board.h
```diff
@@ -0,0 +1,64 @@
...
+// USB UF2
+//--------------------------------------------------------------------+
+
+#define USB_VID           0x239A
+#define USB_PID           0x800B
+#define USB_MANUFACTURER  "ATMegaZero"
+#define USB_PRODUCT       "ATMegaZero ESP32-S2"
+
+#define UF2_PRODUCT_NAME  USB_MANUFACTURER " " USB_PRODUCT
+#define UF2_BOARD_ID      "ATMegaZero-ESP32-S2"
```
hathach, line 60: the board id need to change to match the uf2 spec https://github.com/microsoft/uf2#files-exposed-by-bootloaders

```
```
I will do the manual update.

PR #124, opened 2021-06-23
esp32s2: Add "combined.bin"

jepler:
This is a single file ready to flash at 0x0, similar to the combined.bin for funhouse (except that combined.bin also has an arduino program baked in)


hathach (commented):
I think you also need the ota_data_initial.bin as well for it to boot into tinyuf2 by default.

jepler:
OK, I'll add it!

hathach (approved):
look great, thank for the pr

PR #127, opened 2021-07-06
Refactor imxrt romapi

hathach:
This PR refactor and clean up romapi for nor flash. 
- Except for RT1011, all other mcus has ROM_FLEXSPI_NorFLASH() that we could use https://github.com/NXPmicro/mcux-sdk/blob/main/devices/MIMXRT1062/drivers/fsl_romapi.c for flashing.
- romapi_flash.h/c is addded to abstract that out and implement the ROM_FLEXSPI for rt1011 based on existing code
- romapi_flash.c = bl_romapi_MIMXRT1011.c + flexspi_nor_flash.c to reduce code duplication when adding new rt1011 board

tested and should not affect behavior.


hathach:
merged, any bugs can be fixed as follow up

PR #128, opened 2021-07-07
esp_reset_reason_set_hint() replacement by NVS flag (ESP32-S2)

T94T:
Since `esp_reset_reason_set_hint()` is not working/supported anymore (see: https://github.com/espressif/esp-idf/issues/7230) I decided to rewrite the corresponding code using NVS storage instead.

Please note that this pull request does not work at the moment. There is still a compilation error:
```
tinyuf2/lib/esp-idf/components/nvs_flash/src/nvs_platform.hpp:31:10: fatal error: freertos/FreeRTOS.h: No such file or directory
 #include "freertos/FreeRTOS.h"
          ^~~~~~~~~~~~~~~~~~~~~
```

I hope someone can fix this.


T94T:
will prepare the pull request again.

hathach:
@T94T I dived a bit into this, there is no need for using NVS, the reason hint is still accessible within stage2 bootloader. IDF just slightly changes the reset reason for esp_restart(). I will make an PR to update this.

PR #130, opened 2021-07-09
fix esp32s2 request to uf2 using reset hint

hathach:
reset reason for esp_restart() changed with recent IDF. This fix request to uf2 bootloader. related to #128 

@T94T would you mind trying this out.


T94T:
This seems to work now. Thank you so much for your help!

PR #133, opened 2021-07-15
Added support for my new FeatherS2 Neo

UnexpectedMaker:
Also remembered to dded FeatherS2 Neo to build script this time :)

VID/PID is from Espressif:
https://github.com/espressif/usb-pids/blob/main/allocated-pids.txt

Cheers :)


.github/workflows/build_esp32.yml
```diff
@@ -40,6 +40,7 @@ jobs:
         - 'targett_mcb_wrover'
         - 'unexpectedmaker_feathers2'
         - 'unexpectedmaker_tinys2'
+        - 'unexpectedmaker_feathers2_neo'
```
hathach, line 43: ABC order, should move 1 line up

ports/esp32s2/boards/unexpectedmaker_feathers2_neo/board.h
```diff
@@ -0,0 +1,71 @@
...
+// USB UF2
+//--------------------------------------------------------------------+
+
+#define USB_VID           0x303A
+#define USB_PID           0x80B5
+#define USB_MANUFACTURER  "Unexpected Maker"
+#define USB_PRODUCT       "FeatherS2 Neo"
+
+#define UF2_PRODUCT_NAME  USB_MANUFACTURER " " USB_PRODUCT
+#define UF2_BOARD_ID      "UnexpectedMaker-FeatherS2-Neo"
```
hathach, line 67: board id must respect the uf2 spec https://github.com/microsoft/uf2#files-exposed-by-bootloaders

hathach (changes requested):
Board id must follow the uf2 specs, please correct

```
```

ports/esp32s2/boards/unexpectedmaker_feathers2_neo/board.h
UnexpectedMaker: When did this change of requirement happen?
None of my previous boards required this and most boards in the port don't follow that spec.

Is this then going to be enforced on all previous boards?

.github/workflows/build_esp32.yml
UnexpectedMaker: Sorry, I didn't realise it was alphabetical beyond the starting word.

ports/esp32s2/boards/unexpectedmaker_feathers2_neo/board.h
UnexpectedMaker: Sorry, I stand corrected - most are in this format now, just mine and a few others are not.
Maybe I missed this change of requirements. I'll fix my boards now.

hathach (approved):
great! Thank you for your pr, yeah I try to enforce the board id specs since it is common mistake from user. Sometimes I forgot about it as well.

PR #144, opened 2021-09-05
WIP: Support WeAct mini stm32h7

Bhav97:
I'm trying to add support for [WeAct MiniSTM32H7xx](https://www.adafruit.com/product/5032)

The device has 128K internal flash and 8MB QSPI flash and I cannot figure out how to implement flash routines to accommodate all the flash memories.

Pending
- [ ] board_self_update
- [ ] board_flash_erase_app
- [ ] board_flash_write
- [ ] board_flash_read
- [ ] board_flash_flush
- [ ] board_flash_init
- [ ] board_flash_size
- [ ] board_app_valid

tinyusb enumeration and display works but the image is somewhat misaligned.


hathach:
thank you for your effort along with the pending list. For external flash partition make sure you follow the circuitpython stm32 partition scheme https://github.com/adafruit/circuitpython/tree/main/ports/stm/boards . It will require qspi driver to flash external flash, a new port may take quite a bit of time. I will try to help whenever I could

Bhav97:
I would appreciate any help I can get. 

> For external flash partition make sure you follow the circuitpython stm32 partition scheme https://github.com/adafruit/circuitpython/tree/main/ports/stm/boards

Looking at the circuitpython repo,  I am still quite unsure of what the stm32 partition scheme is and how it would affect tinyuf2. I'm going to try and get the QSPI and SPI memories working first and then continue with the backlog. Ideally I would want all of the memory devices to be available for flash download. (internal + SPI + QSPI)

Bhav97:
I have added SPI and QSPI W25Qx drivers from the MiniSTM32H7xx repository. Still in early stages but you can somewhat load code into the QSPI memory and jump to it. I tried loading a simply blinky program to the QSPI flash. The interrupts seem to break after jumping to the QSPI code.

1. Implemented: `board_flash_init` which initializes the SPI flash and disables memory mapped mode on the QSPI flash to allow writes
2. Implemented: `board_flash_size` returns 8MB. (Even though the total flash size if 8MB+8MB+64KB). But everything works at the moment. 
3. Paritally Implemented: `board_app_valid` checks if the `Reset_Handler` lies in valid flash memory. Stack pointer verification is still pending. 
4. Partially Implemented: `board_flash_read/board_flash_write` maps memory into 3 segments

    * QSPI Flash: `0x9000_0000U - 0x907F_FFFF` (8MB)
    * SPI Flash: `0x6000_0000U - 0x607F_FFFF` (8MB)
    * PFLASH: `0x0801_0000 - 0x0801_FFFF` (64KB)

The initial 64KB is for tinyuf2. Programming the PFLASH is still pending since it would require running off of RAM and I'm struggling with getting the board to load code into RAM.

Bhav97:
Hey @hathach 

Could you check if I'm creating uf2 files right? The code is [here](https://github.com/Bhav97/h750-blinky/blob/11f0bc21fc3a9fd5eccff3b37b86e8dc69c0bff3/CMakeLists.txt#L87)

Bhav97:
I have most things working and implemented now, but there are a few things to test (writing data to SPI flash, self-update and uf2 read). I could test them out along with the review.

The flash driver code isn't very pretty but it works, so I will leave it as is till I can find a portable working replacement. Also the port doesn't handle PFLASH right now (since my board only has the single sector, single bank variant)

elpekenin:
I know there hasn't been any movement out here for 2 years, but i recently bought this board and would like having uf2 on it.

- Is there something missing on the code? I could try and implement myself
- Any idea why the firmware worked a couple days back (screen drawing + storage showing up), but it doesnt do either now?
(i flashed other fimrware via stm32dfu between tests) - Blue LED (E3) does light both when USB is connected and pressing RST button

Waiting for a STLink to get delivered, can't provide further info for now

hathach:
I am not entirely sure as well, PR is still marked as WIP and I kind of forgot about this. @bhav97 do you think this is ready for review and/or merging ?

bhav97:
@hathach , @elpekenin 
Unfortunately the project I was preparing this for didn't work out and I completely forgot about this WIP-PR. 

As for the implementation, all the board_*  callouts are available. I believe we can start a review now.

In general I have 2 concerns for this PR
1.  I created the ports/stm32h7 but right now there's only a single board that uses this variant. There are other variants that have a different sector size and my additions may need some work to make life easier for people who might want to extend tinyuf2 support in the future.

2. The display and QSPI code is straight out of the WeAct repo and is not documented very well
---
> Any idea why the firmware worked a couple days back (screen drawing + storage showing up), but it doesnt do either now?
Flashed other fimrware via stm32dfu between tests
Blue LED (E3) does light both when USB is connected and pressing RST button

I believe stm32dfu doesn't support programming external memory so its possible that tinyuf2 was overwritten. (The MCU variant on the board only has a single flash sector of 128K so its not really possible to program the "internal" flash without overwriting tinyuf2. )
You should be able to load firmware to the "external" QSPI flash and RAM just fine.

elpekenin:
> tinyuf2 was overwritten

Sure, but i mean it doesnt work after flashing the tinyuf2.bin back at it.

bhav97:
That's not good. Does the blue LED stay on? Does double tap bring up the display?

elpekenin:
> That's not good. Does the blue LED stay on? Does double tap bring up the display? 

Can double check later but Led blinked on both events (plugging wire and pressing button), screen did nothing (not even backlight i think)

elpekenin:
Went to the mk file to add `-DRAMCODE` and give that a try, didnt work either.
But thanks to that i realized there is `-flto`, disabled it and everything seems to be working as intended again, weird as heck.
Will now try and compile a `.uf2` and report whether loading user's code works

ports/stm32h7/port.mk
```diff
@@ -0,0 +1,53 @@
+UF2_FAMILY_ID = 0x6db66082 
+CROSS_COMPILE = arm-none-eabi-
+
+ST_HAL_DRIVER = lib/st/stm32h7xx_hal_driver
+ST_CMSIS = lib/st/cmsis_device_h7
+CMSIS_5 = lib/CMSIS_5
+
+# Compiler flags
+CFLAGS += \
+  -flto \
```
elpekenin, line 10: I think LTO is what "broke" code for me (probably gcc-version dependant issue if it worked for you...), do we really need to optimize tinyuf2 code if we aren't using the built-in flash?
Perhaps individual boards using H7xx could opt-in this config, instead of "forcing" its use having it at the MCU level

elpekenin (changes requested):

ports/stm32h7/port.mk
bhav97: That's a good idea, I copied the F4-mkfiles for this port, so there maybe a few more gotchas.
what version of GCC are you compiling with?

elpekenin: 10.3.1 (from ubuntu on WSL2 if that matters)
```
elpekenin@elPeCenin:~$ arm-none-eabi-gcc --version
arm-none-eabi-gcc (15:10.3-2021.07-4) 10.3.1 20210621 (release)
```

---

I cant get the flashed `.uf2` to run, i just get back to bootloader every single time, do you happen to know what im doing wrong here? <https://github.com/elpekenin/qmk_firmware/blob/579e150c72899e78f06e42aefc27d1b71dcaf7ba/keyboards/handwired/onekey/weact_stm32h750vbt6/ld/STM32H750xB_qspi_tinyuf2.ld>. Based this off the `.ld` on ChibiOS and did also take a look on a `.ld` on your GitHub

bhav97: > arm-none-eabi-gcc (15:10.3-2021.07-4) 10.3.1 20210621 (release)

vs 

> Using built-in specs.
COLLECT_GCC=arm-none-eabi-gcc
COLLECT_LTO_WRAPPER=~/Workspace/toolchains/gcc-arm-none-eabi-10.3-2021.10/bin/../lib/gcc/arm-none-eabi/10.3.1/lto-wrapper
Target: arm-none-eabi
Configured with: <removed>
Thread model: single
Supported LTO compression algorithms: zlib
gcc version 10.3.1 20210824 (release) (GNU Arm Embedded Toolchain 10.3-2021.10) 

Definitely seems like something LTO -related is different between our toolchains.

---

 I'm not sure how qmk is laid out in memory, but in your ld, can you try creating 2 memory regions? 
Something like
```
    flash0 : org = 0x90000000, len = 0x300
    flash1 : org = 0x90000300, len = 8m-0x300
```
to explicitly ensure that the vector table comes first

tannewt: The build should output a .map file to show you the final memory layout. The vector table should have its out input section to ensure it is placed correctly.

bhav97:
@hathach  I have rebased over the latest changes, cleaned up the code and history

ports/stm32h7/port.mk
elpekenin: I feel like we should move this to Discord or at least an issue, to stop cluttering the PR (sorry), please answer this message in any way and @ me into it. Unless you think it's better staying here, that is.

The map file looks correct(ish)? Idk, im as noob as you can be with ld stuff....
<https://gist.github.com/elpekenin/95eab3c7831187b86f29b983050e5a4e> (line 2551)

hathach:
> @hathach I have rebased over the latest changes, cleaned up the code and history

superb thank you very much, I will review this as soon as I could. There is no problem with H7 having only this as board. We can refactor this later on when new board is added. Thank you very much for updating the PR.

ports/stm32h7/board_flash.c
```diff
@@ -0,0 +1,228 @@
...
+
+#if BOARD_AXISRAM_EN
+  if (IS_AXISRAM_ADDR(addr) && IS_AXISRAM_ADDR(addr + len - 1))
+  {
+    memcpy(data, (void *) addr, len);
+    return;
+  }
+#endif // BOARD_AXISRAM_EN
+
+  if (IS_PFLASH_ADDR(addr))
```
elpekenin, line 143: Missing SPI check here, only doing QSPI/RAM/InternalFlash
I found this while playing around with the code to try and get my .uf2 to run...
Will try and double-check for similar missing pieces as soon as i get this working...

bhav97: `board_flash_read` is used to provide data for the `CURRENT.uf2`, but it's not really possible to target multiple memories for a readout. (maybe if we readout internal/QSPI/SPI flash and then concatenate all data into CURRENT.uf2)
Internally within the port it's used to read the vector table, and the SPI flash cannot contain executables, so I left it out of `board_flash_read`.

hathach:
@bhav97 if you are busy, I could make the review changes myself so that we could finally merge this :)

bhav97:
@hathach I have made the requested changes but the underlying issue from #342 is still present. I'll open another PR when I figure it out.

Thank you :)

hathach (approved):
perfect, thank you very much for your work and patient. No worry about the other issue, this PR already done an amazing lots of works. We can always fix it later, and/or add support for on-flash, refactor etc...

PR #148, opened 2021-09-10
Add TCA9554 peripheral power control for neopixel support

skieast:
Added support for onboard neopixel that has its power control via the TCA9554

### DIFF
--- a/ports/esp32s2/boards/boards.c
+++ b/ports/esp32s2/boards/boards.c
@@ -60,6 +60,11 @@ void dotstar_init(void);
 void dotstar_write(uint8_t const rgb[]);
 #endif
 
+#ifdef TCA9554_ADDR
+#include "esp_err.h"
+#include "driver/i2c.h"
+#endif
+
 #ifdef LED_PIN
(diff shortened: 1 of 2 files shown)


hathach (commented):
thank you for your PR, I don't have this board to test with, but it looks good. `IR_RGB_CTRL` is connected to P6 of TCA9554. Though i have a question during the boot2 stage, it will try to turn on the neopixel purple to indicate double reset, would that work as well (I guess not).

image

skieast:
True, the PERI_POWER is not enabled during boot2.  I may look at that but am more interested in getting the display working. My test board display has actually stopped working so NEOPIXEL came to the front. Thanks for looking at this.

hathach:
> True, the PERI_POWER is not enabled during boot2. I may look at that but am more interested in getting the display working. My test board display has actually stopped working so NEOPIXEL came to the front. Thanks for looking at this.

thank for confirmation with boot2, this is still good to merge.

hathach (approved):
thank you for your PR, look good

PR #152, opened 2021-09-30
stm32l4 support for swan_r5

m-mcgowan:
Tested and is working but will need more thorough testing.


m-mcgowan:
Didn't intend to create the PR here!  We do want to make a PR eventually, but not on the initial commit! I'll re-create the PR in our own fork.

hathach:
no problem at all, for the L3 uf2 family id, it can be found in uf2 repo
https://github.com/microsoft/uf2/blob/master/utils/uf2families.json#L168

m-mcgowan:
Thanks for the pointer! I'll update the ID.

PR #158, opened 2021-11-20
fix board_flash_write() on stm32f4

oliver-joos:
Bug symptom: if an image for stm32f4 begins with >=256 bytes that already match flash content then the image must be written twice.
Reason: if the first flash_write() into an erase block matches flash content, the erase is delayed and will destroy that matching content later.


hathach (approved):
you are right, stm32 flash doesn't implement sector caching like other port. I probably got mixed up with other ports. Thank you very much for your fix

PR #161, opened 2021-12-05
add support for stm32f412

oliver-joos:
This branch has been tested successfully on a Nucleo F412ZG board (144 pins!) and on custom boards with F412CE (only 48 pins). It will also help to support boards with F413, F423, F446, F469 and F479.


hathach (approved):
thanks, it looks good. We probably need to move vbus setup to board.h in the future since it is different from boards to boards.

PR #162, opened 2021-12-05
accept custom BOARD_DIR or custom APP_START vector for stm32

oliver-joos:
A custom BOARD_DIR allows to use tinyuf2 for example as Git submodule in another project that has its own board.h and board.mk.
To set the (absolute or relative) path to your board directory just run:
`make BOARD=<my-board> BOARD_DIR=<path/of/my-board/> ...`

A custom APP_START allows to use tinyuf2 for applications with other start vectors than the hard-coded default in tinyuf2.
To set the start vector of your application just run:
`make APP_START=0x08020000 ...`
or define it in your "board.h" with:
`#define BOARD_FLASH_APP_START 0x08020000`

### DIFF
--- a/ports/lpc55/port.mk
+++ b/ports/lpc55/port.mk
@@ -37,7 +37,7 @@ SRC_C += \
 # Port include
 INC += \
   $(TOP)/$(PORT_DIR) \
-  $(TOP)/$(BOARD_DIR) \
+  $(BOARD_DIR) \
 	$(TOP)/$(SDK_DIR)/CMSIS/Include \
 	$(TOP)/$(MCU_DIR) \
 	$(TOP)/$(MCU_DIR)/project_template \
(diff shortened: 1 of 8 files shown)


hathach (commented):
Thank you for your PR though
1. tinyuf2's `internal` makefile will continue to change and may not work with other project. It would be better that your application have its own makefile (by duplicating and modifying tinyuf2)
2. `BOARD_FLASH_APP_START` can be defined per board within its board.h . We don't really need to introduce the APP_START, just wrap the BOARD_FLASH_APP_START with ifndef is good enough. This should be done in its own PR since (1) is not accepted.

oliver-joos:
Sorry for the long delay! Finally I had time to upgrade my project using tinyuf2 0.9.2.

You are right about (1). I can use `make BOARD_DIR="<path>"`while I don't like that `<path>` has to be relative to $(TOP) of tinyuf2. I tried to solve this with PR #162. But I use tinyuf2 as a git submodule in my project, so its relative path never changes ⇒ problem solved.

As for (2) I opened PR #212.

oliver-joos:
This PR is replaced by my PR #212

Issue #165, opened 2021-12-11
STM32F4 is not building for gcc-11


DuMaM:
### Operating System

Linux

### INFO_UF2.TXT

There is no need of info_uf2.txt file.

### What happened ?

I'm unable to build current release for stm32f4, also there is no update img in github release.

### How to reproduce ?

1. Make sure you are up to date with master branch and submodules
2. Check if your gcc version is:
``` shell
arm-none-eabi-gcc (Arch Repository) 11.2.0
Copyright (C) 2021 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```
3. Go to ports/stm32f4
3. Run `make BOARD=stm32f411ce_blackpill all`

### Debug Log

```In function 'board_usb_get_serial',
    inlined from 'tud_descriptor_string_cb' at /home/dumam/Projects/tinyuf2/src/usb_descriptors.c:190:20,
    inlined from 'process_get_descriptor' at /home/dumam/Projects/tinyuf2/lib/tinyusb/src/device/usbd.c:984:50,
    inlined from 'process_control_request' at /home/dumam/Projects/tinyuf2/lib/tinyusb/src/device/usbd.c:680:11,
    inlined from 'tud_task' at /home/dumam/Projects/tinyuf2/lib/tinyusb/src/device/usbd.c:522:15,
    inlined from 'tud_task' at /home/dumam/Projects/tinyuf2/lib/tinyusb/src/device/usbd.c:474:6,
    inlined from 'main' at /home/dumam/Projects/tinyuf2/src/main.c:92:5:
/home/dumam/Projects/tinyuf2/ports/stm32f4/boards.c:241:3: error: 'memcpy' reading 12 bytes from a region of size 0 [-Werror=stringop-overread]
  241 |   memcpy(serial_id, STM32_UUID, len);
      |   ^
compilation terminated due to -Wfatal-errors.
lto1: all warnings being treated as errors
lto-wrapper: fatal error: arm-none-eabi-gcc returned 1 exit status
compilation terminated.
/usr/lib/gcc/arm-none-eabi/11.2.0/../../../../arm-none-eabi/bin/ld: error: lto-wrapper failed
collect2: error: ld returned 1 exit status
make: *** [../rules.mk:44: _build/stm32f411ce_blackpill/tinyuf2-stm32f411ce_blackpill.elf] Błąd 1
[...]

PR #166, opened 2021-12-11
Build fix for gcc 11 on stm32 boards

DuMaM:
This is a fix for issue #165


ports/stm32f3/boards.c
```diff
@@ -30,7 +30,7 @@
 // MACRO TYPEDEF CONSTANT ENUM DECLARATION
 //--------------------------------------------------------------------+
 
-#define STM32_UUID ((uint32_t *)0x1FFFF7AC)
+#define STM32_UUID ((uint32_t *)0x1FFFF7ACUL)
```
hathach, line 33: can you try to add volatile as follow and keep memcpy to see if that make gcc11 happy 

```C
#define STM32_UUID ((volatile uint32_t *)0x1FFFF7AC)
```

DuMaM: I tried this already and this don't work.

ports/stm32f3/boards.c
```diff
@@ -190,7 +190,11 @@ void board_app_jump(void)
 uint8_t board_usb_get_serial(uint8_t serial_id[16])
 {
   uint8_t const len = 12;
-  memcpy(serial_id, STM32_UUID, len);
+  uint32_t word;
+  for (int i = 0; i < len; i=i+4) {
+    word = READ_REG(*(STM32_UUID+i));
```
hathach, line 195: thanks for above confirmation, the expression is not correct `STM32_UUID+i` with i = 4n is not correct since STM32_UUID is already declared as uint32_t*. I will try to install gcc11 and try later.

DuMaM: oh thats true. Good catch.

hathach (commented):
I couldn't reproduce the issue since latest release from ARM is only 10.3. We will leave this issue and pr pending until gcc11 is released.  Thank you for the pr.

hathach:
thank you for keeping this PR update, I did pull and try to update this PR as well. Unfortunately the self-update target of stm32f4 still failed due to the false warnings from -Werror=stringop-overread gcc. This has been fixed in gcc 11.3. For now, we will suppress it with -Wno-error to get this merged asap.

https://gcc.gnu.org/git/gitweb.cgi?p=gcc.git;h=91f7d7e1bb6827bf8e0b7ba7eb949953a5b1bd18

PR #174, opened 2021-12-25
extract SDP_WRITE_ADDR (_fcfb_origin) and SDP_JUMP_ADDR (_ivt_origin) from linker file

hathach:
- use `sed` to extract SDP_WRITE_ADDR (_fcfb_origin) and SDP_JUMP_ADDR (_ivt_origin) from linker file. 
- This helps to reduce the amount of manual work and magic number should we make changes to linker file.

### DIFF
--- a/ports/mimxrt10xx/Makefile
+++ b/ports/mimxrt10xx/Makefile
@@ -25,43 +25,39 @@ include ../rules.mk
 #---------------------------------------------------------
 # Load to SRAM using sdphost
 # Note: you may need to give the sdphost binary executable permission first.
+#
 # SDP loads the image into the RAM locations specified in the .ld files.
-# SDP Write Address must equal _fcfb_origin value
-# SDP Jump Address must equal _ivt_origin = _fcfb_origin + _fcfb_length
-# TinyUF2 will copy itself to the correct location in flash
-# UF2 Write Address shows where the image will reside in flash if you 
+#  - "SDP Write Address" must equal _fcfb_origin
+#  - "SDP Jump Address" must equal _ivt_origin
+#
+# TinyUF2 will copy itself to the correct location in flash.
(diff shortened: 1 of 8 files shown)


nxf58843:
Sorry for the delayed response.  There are some issues with this update.  The image runs after loaded but does not run from flash (RT1010 EVK).  Let me try to clean this up.  This PR should probably be reverted.

nxf58843:
I think it would also be good to switch to spsdk (https://pypi.org/project/spsdk/) for loading, then we do not need to keep the sdphost binaries.  They can load it separately like pyocd.

hathach:
> Sorry for the delayed response. There are some issues with this update. The image runs after loaded but does not run from flash (RT1010 EVK). Let me try to clean this up. This PR should probably be reverted.

I tested with rt1010 evk and seems to work well. This PR essentially does not make any changes in the compile binary. It only add the sed script to extract the **_fcfb_origin and _ivt_origin** from the linker file. Which OS you are using, I haven't tested on windows, maybe adding this line to makefile to check if sed does its work correctly

```
$(error Write: $(SDP_WRITE_ADDR) Jump: $(SDP_JUMP_ADDR) )
```

> I think it would also be good to switch to spsdk (https://pypi.org/project/spsdk/) for loading, then we do not need to keep the sdphost binaries. They can load it separately like pyocd.

Yeah, I agree, we could switch to spsdk in the future. If possible, please make an PR for it. I could do it as well if you don't have time.

gsteiert:
I forgot that the RT1010 does not fit with debug logs enabled.  I think that was my problem.  It is working for me now.  I will submit a pull request with the SPSDK update.  It is mostly documentation, let me know if anything is unclear.

hathach:
thank you for confirmation :)

PR #175, opened 2021-12-25
fix red LED pin

ladyada:
@hathach do you know why pin 42 might not be working for RC? its a JTAG pin, not sure if its special...


hathach:
> @hathach do you know why pin 42 might not be working for RC? its a JTAG pin, not sure if its special...

Yeah, I remembered testing this. JTAG pins couldn't be used for double tap RC. It is probably got special init on power/reset.

hathach (approved):

PR #178, opened 2021-12-29
NXP SPSDK for MIMXRT builds

gsteiert:
This pull request removes the OS checking for sdphost and instructs the user to install the NXP SPSDK from PyPI.
The directory with with sdphost binaries has a note that it is being depricated and where to find SPSDK.
Most of the edits are in mimxrt10xx/README.md 
Let me know if anything is unclear.


hathach (approved):
look great, thank you for the update. We will leave the existing sdphost binaries/source for reference or just in case the spsdk has any compatible issue.

PR #181, opened 2022-01-05
fix double tap issue

### LINKED ISSUE #176: PIN_DOUBLE_RESET_RC on ESP32-S3 - using IO47 not working
### Operating System

MacOS

### INFO_UF2.TXT

TinyUF2 Bootloader 0.6.2-29-g06a65c8 - esp-idf (v4.4-dev-3608-gbbe2a1bf34) tinyusb (0.12.0-203-ga4cfd1c6)
Model: Unexpected Maker TinyS3
Board-ID: ESP32S3-TinyS3-01
Date: Dec 27 2021

### What happened ?

I'm using IO47 for my PIN_DOUBLE_RESET_RC and it's not working and I'm not sure why.

Datasheet shows that it's Reset mode is E1 (Input enabled) but it's default function is (F0) - SPICLK_P_DIFF which is O/T, not GPIO47 which is I/O/T - (F1)

I've confirmed I have plenty of hold high time on the IO during the boot cycle.  I've even gone to a 2.2uF cap with 1Mohm resistor, and I can see it's holding 3V+ for ages - way longer than I need... but for whatever reason, it never catches it as high and just moves on to load into CPY.

I've confirmed it running through the correct code as I have the LED going blue instead of purple if PIN_DOUBLE_RESET_RC code is being run, and everything looks correct.

I can't see anything in the TinyUF2 or IDF code that would prevent IO47/48 from being used as a GPIO - they are both new and exclusive to the S3, but the only difference is the F0/F1 states.

I've even tried to brute force it...
```
esp_rom_gpio_pad_select_gpio(PIN_DOUBLE_RESET_RC);
gpio_ll_output_disable(&GPIO, PIN_DOUBLE_RESET_RC);
gpio_ll_input_enable(&GPIO, PIN_DOUBLE_RESET_RC);
```

I don't have access to `driver/gpio.h` in `bootloader-start.c` so I was unable to try `gpio_config` to set the GPIO to Input.

Any ideas what I can do?   IO47/48 are the only IO I have spare that I can use for PIN_DOUBLE_RESET_RC so I need to get this working.

Thanks!

### How to reproduce ?

Not sure how you can repo this.

### Debug Log

_No response_

### Screenshots

_No response_

hathach:
fix double tap issue, fix #176 . Root cause: `gpio_ll_set_level()` is used in version 0.7.0 since S3 gpio rom does not have `gpio_output_set_high()`, unlike rom API, gpio_ll needs to explicitly enable output first.

@ladyada while working on this issue, I figured out that `esp_rom_gpio_pad_select_gpio()` can also be used to get JTAG pin working as double tap pin. Therefore there is no restriction on JTAG pin after this got merged. Please skip 0.7.0 for new S2 board, 0.8.0 will be released as soon as this is merged.

### DIFF
--- a/ports/espressif/components/bootloader/subproject/main/bootloader_start.c
+++ b/ports/espressif/components/bootloader/subproject/main/bootloader_start.c
@@ -212,6 +212,7 @@ static int selected_boot_partition(const bootloader_state_t *bs)
 
 #ifdef PIN_DOUBLE_RESET_RC
           // Double reset detect if board implements 1-bit memory with RC components
+          esp_rom_gpio_pad_select_gpio(PIN_DOUBLE_RESET_RC);
           PIN_INPUT_ENABLE(GPIO_PIN_MUX_REG[PIN_DOUBLE_RESET_RC]);
           if ( gpio_ll_get_level(&GPIO, PIN_DOUBLE_RESET_RC) == 1 )
           {
@@ -220,6 +221,7 @@ static int selected_boot_partition(const bootloader_state_t *bs)
           }
           else
           {
+            gpio_ll_output_enable(&GPIO, PIN_DOUBLE_RESET_RC);
             gpio_ll_set_level(&GPIO, PIN_DOUBLE_RESET_RC, 1);
           }
 #endif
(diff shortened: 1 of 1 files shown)


ladyada:
ok great work!

UnexpectedMaker:
Well done!!!!  Can you wait on 0.8.0 for 10 mins while I do a PR for my new S3 boards please?

PR #182, opened 2022-01-18
add lil s3

ladyada:
QT Py S3 with 4MB flash internal
with RC on pin 10, neopixel on 38


hathach (approved):
tested and working with actual prototype board

PR #185, opened 2022-01-19
Add favicon icon support

hathach:
- favicon icon + autorun only work with windows, but nice to have
- feature can be enable per board by adding `#define TINYUF2_FAVICON 1`. Enabled on esp32s2 saola 1 wrover as example
- add adafruit logo as `favicon.h` (256x256 with only 7.5KB) as example for now. We could have a final per-project (tinyuf2 logo) or even per-board favicon later if needed.

@ladyada This PR took longer than expected since we need to update and extend the ghostfat to support static files with size larger than 1 sector/cluster which is the limit previously for simple calculation for FAT indexing.

@henrygab If you have time, could you help to review the changes. I try the best I could, but you have far more knowledge on FAT FS than me. Not a problem at all if you are busy, we could fix issue with follow up PR later on.

screenshot on my windows VM

image

image


src/ghostfat.c
```diff
@@ -38,6 +38,12 @@
 
 #define STATIC_ASSERT(_exp) _Static_assert(_exp, "static assert failed")
 
+#define STR0(x) #x
+#define STR(x) STR0(x)
+
+#define UF2_ARRAY_SIZE(_arr)    ( sizeof(_arr) / sizeof(_arr[0]) )
```
henrygab, line 44: This is not a typesafe macro, and has historically been the source of many really troublesome bugs.
I strongly recommend using a typesafe alternative.

<details><summary>Recommended alternative definition</summary><P/>

```C++

#if defined(__cplusplus)
     /*
        Works on older compilers, even Visual C++ 6....
        Created by Ivan J. Johnson, March 06, 2007
        See http://drdobbs.com/cpp/197800525?pgno=1
        Full description is in markdown file array_size2.md
    */
    #define ARRAY_SIZE2(arr) ( \
       0 * sizeof(reinterpret_cast<const ::Bad_arg_to_ARRAY_SIZE2*>(arr)) + /*check1*/ \
       0 * sizeof(::Bad_arg_to_ARRAY_SIZE2::check_type((arr), &(arr)))    + /*check2*/ \
       sizeof(arr) / sizeof((arr)[0])                                       /* eval */ \
       )

    struct Bad_arg_to_ARRAY_SIZE2 {
       class Is_pointer; // incomplete
       class Is_array {};
       template <typename T>
       static Is_pointer check_type(T const *, T const * const *);
       static Is_array check_type(void const *, void const *);
    };
#elif !defined(__cplusplus) && defined(__GNUC__)

    // Even C can have type-safety for equivalent of ARRAY_SIZE() macro,
    // when using the following two GCC extensions:
    //     typeof()
    //     __builtin_types_compatible_p()

    #if defined(ARRAYSIZE2_SHOW_VERSION_MESSAGE)
        #pragma message( "ARRAY_SIZE2 -- Using GNUC version" )
    #endif

    // validated using:
    //   MSP430  gcc   4.5.3
    //   x86-64  icc  16.0.3
    //   x86-64  gcc   4.1.2
    //   x86-64 clang  3.0.0
    //   AVR     gcc   4.5.4
    //   ARM     gcc   4.5.4

    #define __SIMPLEHACKS_COMPATIBLE_TYPES__(a,b)   __builtin_types_compatible_p(typeof(a), typeof(b)) // GCC extensions
    #define __SIMPLEHACKS_BUILD_ERROR_ON_NONZERO__(x)  (sizeof(struct { int:-!!(x)*0x1ee7;})) // if x is zero, reports "error: negative width in bit-field '<anonymous>'"
    #define __SIMPLEHACKS_MUST_BE_ARRAY__(x)        __SIMPLEHACKS_BUILD_ERROR_ON_NONZERO__(__SIMPLEHACKS_COMPATIBLE_TYPES__((x), &(*x)))
    #define ARRAY_SIZE2(_arr)       ( (sizeof(_arr) / sizeof((_arr)[0])) + __SIMPLEHACKS_MUST_BE_ARRAY__(_arr) ) // compile-time error if not an array

```

</details><HR/>

Better yet, just use the header from https://github.com/SimpleHacks/UtilHeaders/ (MIT Licensed).

src/ghostfat.c
```diff
@@ -98,10 +105,8 @@ struct TextFile {
 #define FAT_ENTRY_SIZE            (2)
 #define FAT_ENTRIES_PER_SECTOR    (BPB_SECTOR_SIZE / FAT_ENTRY_SIZE)
 // NOTE: MS specification explicitly allows FAT to be larger than necessary
-#define TOTAL_CLUSTERS_ROUND_UP   ( (BPB_TOTAL_SECTORS / BPB_SECTORS_PER_CLUSTER) + \
-                                   ((BPB_TOTAL_SECTORS % BPB_SECTORS_PER_CLUSTER) ? 1 : 0))
-#define BPB_SECTORS_PER_FAT       ( (TOTAL_CLUSTERS_ROUND_UP / FAT_ENTRIES_PER_SECTOR) + \
-                                   ((TOTAL_CLUSTERS_ROUND_UP % FAT_ENTRIES_PER_SECTOR) ? 1 : 0))
+#define TOTAL_CLUSTERS_ROUND_UP   UF2_DIV_CEIL(BPB_TOTAL_SECTORS, BPB_SECTORS_PER_CLUSTER)
+#define BPB_SECTORS_PER_FAT       UF2_DIV_CEIL(TOTAL_CLUSTERS_ROUND_UP, FAT_ENTRIES_PER_SECTOR)
```
henrygab, line 109: The use of the `UF2_DIV_CEIL` macro makes this much easier to read.  Well done.

src/ghostfat.c
```diff
@@ -266,31 +306,49 @@ void uf2_read_block (uint32_t block_no, uint8_t *data)
...
+      // cluster 1 is reserved
+      data16[1] = 0xffff;
+
       // WARNING -- code presumes only one NULL .content for .UF2 file
-      //            and all non-NULL .content fit in one sector
       //            and requires it be the last element of the array
-      uint32_t const end = (NUM_FILES * FAT_ENTRY_SIZE) + (2 * FAT_ENTRY_SIZE);
-      for ( uint32_t i = 1; i < end; ++i ) data[i] = 0xff;
+      uint32_t v = 2;
+      for (uint32_t i=0; i<NUM_FILES-1; i++)
```
henrygab, line 323: Minor nit: use `++i` instead of `i++`

src/ghostfat.c
```diff
@@ -136,20 +136,26 @@ const char indexFile[] =
...
+static TextFile_t const info[] = {
+    {.name = "INFO_UF2TXT", .content = infoUf2File, .size = sizeof(infoUf2File) - 1},
+    {.name = "INDEX   HTM", .content = indexFile  , .size = sizeof(indexFile  ) - 1},
+#if TINYUF2_FAVICON
+    {.name = "AUTORUN INF", .content = autorunFile, .size = sizeof(autorunFile) - 1},
+    {.name = "FAVICON ICO", .content = (char const *) favicon_data, .size = favicon_len},
+#endif
     // current.uf2 must be the last element and its content must be NULL
-    {.name = "CURRENT UF2", .content = NULL},
+    {.name = "CURRENT UF2", .content = NULL       , .size = 0},
```
henrygab, line 152: I'm recommending later that the `size` field `CURRENT.UF2` should be filled in.  See later for why

src/ghostfat.c
henrygab, line 323: This is identical to the code for the `CURRENT.UF2` file.  Both this and the following loop are both filling in the FAT chain for a given file entry.  In all cases, the file contents are stored contiguously.

So, rather than looping until `i < NUM_FILES-1`, this would loop until `i < NUM_FILES`.

If the file size field is correct for CURRENT.UF2, then this should "just work".

src/ghostfat.c
```diff
@@ -338,20 +396,27 @@ void uf2_read_block (uint32_t block_no, uint8_t *data)
...
+    // plus 2 for first data cluster offset
+    uint32_t const cluster_num = 2 + sectionRelativeSector / BPB_SECTORS_PER_CLUSTER;
+
+    if ( cluster_num < UF2_FIRST_CLUSTER_NUMBER )
     {
-      // WARNING -- code presumes first data cluster == first file, second data cluster == second file, etc.
-      struct TextFile const * inf = &info[ sectionRelativeSector / BPB_SECTORS_PER_CLUSTER ];
-      size_t fileContentStartOffset = (sectionRelativeSector % BPB_SECTORS_PER_CLUSTER) * BPB_SECTOR_SIZE;
-      size_t fileContentLength = strlen(inf->content);
+      // Files that is not CURRENT.UF2
```
henrygab, line 411: grammar nit pick: choose one of:

`// Files that are not CURRENT.UF2"`

`// File that is not "CURRENT.UF2"`

I think the intent is:

`// Handle all files that are not "CURRENT.UF2"`

src/ghostfat.c
```diff
@@ -338,20 +396,27 @@ void uf2_read_block (uint32_t block_no, uint8_t *data)
...
+      // first data cluster == first file
+      uint32_t fid = info_index_of(cluster_num);
```
henrygab, line 413: I am working on the presumption that you have thoroughly unit-tested both `info_index_of()` and `info_cluster_start()`.   With that in mind, these are great abstractions, which help make the code readable.  Well done!

src/ghostfat.c
```diff
@@ -368,8 +433,8 @@ void uf2_read_block (uint32_t block_no, uint8_t *data)
     }
     else
     {
-      // generate the UF2 file data on-the-fly
-      sectionRelativeSector -= (NUM_FILES - 1) * BPB_SECTORS_PER_CLUSTER;
+      // CURRENT.UF2: generate data on-the-fly
+      sectionRelativeSector -= (UF2_FIRST_CLUSTER_NUMBER-2) * BPB_SECTORS_PER_CLUSTER;
```
henrygab, line 437: why not `info_cluster_start(NUM_FILES-1)`?
It would understand the sameness of these two code paths.

henrygab (approved):
At first glance, this looks good.  Lots of nice abstractions to make the code more readable.

**_I am presuming that both `info_index_of()` and `info_cluster_start()` were unit-tested._**  That is, that you've sanity-checked those functions give valid outputs for all valid inputs.

I have some recommendations and questions (see comments).

henrygab (commented):
Should `favicon.h` be in the common source directory, or should it instead be created in each board's subdirectory?

I think since the icon should be customizable on a per-board basis, it should be placed into each board's directory.  Or, at the minimum, provide a way to override the default favicon.

src/ghostfat.c
henrygab: I am making this change in a branch, so you can review easily.

henrygab:
one more change will be necessary with this checkin:

Generation of the "known good" images for ghostfat self-tests.  I am working on this in a branch.  I have hand-validated `test_base_with_favicon` (I may change the names to be shorter).

* [x] test_base_with_favicon
* [ ] test_4k_cluster_with_favicon
* [ ] test_32k_cluster_with_favicon

src/ghostfat.c
henrygab: fixed in my branch

henrygab, line 323: Two issues exist in the code as written:

1. Assumption that FAT chain for all files (other than CURRENT.UF2) will be in first sector of FAT
2. No `static_assert` to ensure that this condition is provably true.

henrygab (changes requested):
Because the new code allows for >1 sector per file, there is now higher risk of files having FAT chain beyond the first sector of each FAT. 

I will look at either (1) adding necessary `static_assert` to ensure this is avoided, or (2) fixing the FAT chain code to work always.

src/ghostfat.c
henrygab: fixing in my branch of this ... will provide PR when it's ready

src/ghostfat.c
henrygab: fixed in my branch

src/ghostfat.c
hathach: Yeah, I am aware of this. It is only dangerous when using with dynamic pointer/reference. Since we only use it for getting the number of static files locally. It is rather safe, though I will update it to declare NUM_FILES as enum later on.

src/ghostfat.c
hathach: Thank you

src/ghostfat.c
hathach: Yeah, I agree, that makes it more consistent

src/ghostfat.c
hathach: I saw this in the current code, though in my testing, the 0xffff for CURRENT.UF2 (NUM_FILES-1) is overwritten anyway later on, therefore I removed it.

src/ghostfat.c
hathach: Right, I am terrible with grammar

src/ghostfat.c
hathach: Yeah, I know ++i is better, but it is into my habit already.

src/ghostfat.c
hathach: Yeah, totally agreee

hathach:
> Should `favicon.h` be in the common source directory, or should it instead be created in each board's subdirectory?
> 
> I think since the icon should be customizable on a per-board basis, it should be placed into each board's directory. Or, at the minimum, provide a way to override the default favicon.

This is just initial implementation. There would be an project-wide icon which is in the src directory. The board-specific icon will be supported as well. I am thinking to have TINYUF2_FAVICON as a string default to `favicon.h`, which can be changed by board.h to other name e.g `my_icon.h` and placed in the board folder itself. 

> one more change will be necessary with this checkin:
> Generation of the "known good" images for ghostfat self-tests. I am working on this in a branch. I have hand-validated test_base_with_favicon (I may change the names to be shorter).

I am looking forward to your PR.

hathach:
> Because the new code allows for >1 sector per file, there is now higher risk of files having FAT chain beyond the first sector of each FAT.
> 
> I will look at either (1) adding necessary `static_assert` to ensure this is avoided, or (2) fixing the FAT chain code to work always.

You are right, I will update PR to include the check/warning or fix the code if I could do it quickly. I am off until next week, will do this in the next week.

Btw if you are making changes to your branch, would you mind make an PR toward this `add-favicon`, so that you don’t have to hold/wait for this PR to be merged. We could just merge your PR to this branch, and this PR will be updated.

Thank you very much for the thoroughly review. I really appreciate your feedback.

henrygab:
I'm just working (slowly) through the last bugs in the new FAT generation.
I will make a PR when it's done.   I am fighting (long?) COVID, so have fewer hours awake.
If I have time, I will add ability to override favorite icon also.

henrygab:
@hathach -- Please see PR #187, which fixes things and adds additional test images.

src/ghostfat.c
henrygab: resolved in PR #187

hathach:
> @hathach -- Please see PR #187, which fixes things and adds additional test images.

thank you very much, that PR is brilliant and merged. I will do a bit more tweak to the PR and we could wrap this up.

hathach:
@henrygab I pushed more "enhancement" to ghostfat, please review if you have time and see if that makes sense to you https://github.com/adafruit/tinyuf2/pull/185/commits/252151a18e8e21c8f81ce4f1bd019aef1f84a3ae

src/ghostfat.c
```diff
@@ -33,9 +33,9 @@
 #include "board_api.h"
 #include "uf2.h"
 
-// use as: GF_DEBUG_PRINT(("foo %d", x));
+// use as: GF_DEBUG_PRINT("foo %d", x);
 #if defined(DEBUG)
-  #define GF_DEBUG_PRINT(x) do { printf x; fflush(stdout); } while (0)
+  #define GF_DEBUG_PRINT(x) do { printf(x); fflush(stdout); } while (0)
```
henrygab, line 38: Similarly, the lack of parenthesis here was also intentional.

src/ghostfat.c
```diff
@@ -174,8 +175,11 @@ static FileContent_t const info[] = {
     {.name = "CURRENT UF2", .content = NULL       , .size = 0                       },
 };
 
+enum {
+  NUM_FILES = sizeof(info) / sizeof(info[0]),
```
henrygab, line 179: Having this in an enum does not prevent bugs from refactoring.  All it does is ensure the result is (essentially) `constexpr`.  If `info` is later refactored to be a pointer, the code breaks.

Using this style of unsafe `ARRAY_SIZE` code is one of the entrenched "bad practices" that I try to help people realize are dangerous.  Please consider using a safe(r) version.

src/ghostfat.c
```diff
@@ -339,16 +315,16 @@ void uf2_read_block (uint32_t block_no, uint8_t *data)
   memset(data, 0, BPB_SECTOR_SIZE);
   uint32_t sectionRelativeSector = block_no;
 
-  if ( block_no == 0 ) // aka requesting boot block
+  if ( block_no == 0 )
   {
-    // Requested boot block
+    // Request Boot block
```
henrygab, line 316: The old word "Requested" was past-tense, and referred to the command that was received.

Changing these to "Request" (present-tense) doesn't make sense, because the code is not making a request.

If you want to change the tense from past to present, perhaps you meant "create" or "generate"?

src/ghostfat.c
```diff
@@ -441,38 +416,35 @@ void uf2_read_block (uint32_t block_no, uint8_t *data)
...
+    // aka read from the data area (files, unused space, ...)
     sectionRelativeSector -= FS_START_CLUSTERS_SECTOR;
 
     // plus 2 for first data cluster offset
-    uint32_t const cluster_num = 2 + sectionRelativeSector / BPB_SECTORS_PER_CLUSTER;
+    uint32_t fid = info_index_of(2 + sectionRelativeSector / BPB_SECTORS_PER_CLUSTER);
+    FileContent_t const * inf = &info[fid];
 
-    if ( cluster_num < UF2_FIRST_CLUSTER_NUMBER )
+    sectionRelativeSector -= (info[fid].cluster_start-2) * BPB_SECTORS_PER_CLUSTER;
```
henrygab, line 439: Don't re-use `sectionRelativeSector` ... create new variable, such as `fileRelativeSector`.  The compiler will optimize it anyways, but it will be much clearer what value is stored....

henrygab (commented):
Initial review of this commit

src/ghostfat.c
hathach: we could just use the `VA_ARGS` macros instead

src/ghostfat.c
```diff
@@ -178,15 +178,11 @@ static FileContent_t info[] = {
...
 };
 
-#define NUM_DIRENTRIES     (NUM_FILES + 1) // Code adds volume label as first root directory entry
-#define REQUIRED_ROOT_DIRECTORY_SECTORS UF2_DIV_CEIL(NUM_DIRENTRIES+1, DIRENTRIES_PER_SECTOR)
-
-STATIC_ASSERT(ROOT_DIR_SECTOR_COUNT >= REQUIRED_ROOT_DIRECTORY_SECTORS);         // FAT requirement -- Ensures BPB reserves sufficient entries for all files
-STATIC_ASSERT(NUM_DIRENTRIES < (DIRENTRIES_PER_SECTOR * ROOT_DIR_SECTOR_COUNT)); // FAT requirement -- end directory with unused entry
-STATIC_ASSERT(NUM_DIRENTRIES < BPB_ROOT_DIR_ENTRIES);                            // FAT requirement -- Ensures BPB reserves sufficient entries for all files
-STATIC_ASSERT(NUM_DIRENTRIES < DIRENTRIES_PER_SECTOR); // GhostFAT bug workaround -- else, code overflows buffer
+STATIC_ASSERT(NUM_DIRENTRIES <= BPB_ROOT_DIR_ENTRIES);  // FAT requirement -- Ensures BPB reserves sufficient entries for all files
```
henrygab, line 184: Because FAT requires the final direntry to be unused, should this be `<` instead of `<=`?

src/ghostfat.c
```diff
@@ -178,15 +178,11 @@ static FileContent_t info[] = {
...
+STATIC_ASSERT(NUM_DIRENTRIES <= DIRENTRIES_PER_SECTOR); // GhostFAT bug workaround -- else, code overflows buffer
```
henrygab, line 185: Because FAT requires the final direntry to be unused, should this be `<` instead of `<=`?

henrygab (approved):
Overall, you've made the code better.  I do think there are a couple errors with the new static assertions, but should be good, especially if the self-tests pass.

src/ghostfat.c
henrygab: I just went ahead and fixed this.

src/ghostfat.c
henrygab: I just went ahead and fixed this.

src/ghostfat.c
```diff
@@ -436,12 +436,12 @@ void uf2_read_block (uint32_t block_no, uint8_t *data)
     uint32_t fid = info_index_of(2 + sectionRelativeSector / BPB_SECTORS_PER_CLUSTER);
     FileContent_t const * inf = &info[fid];
 
-    sectionRelativeSector -= (info[fid].cluster_start-2) * BPB_SECTORS_PER_CLUSTER;
+    uint32_t fileRelativeSector -= (info[fid].cluster_start-2) * BPB_SECTORS_PER_CLUSTER;
```
hathach, line 439: should be 
```C
uint32_t fileRelativeSector = sectionRelativeSector - (info[fid].cluster_start-2) * BPB_SECTORS_PER_CLUSTER;
```

src/ghostfat.c
hathach: thank you

src/ghostfat.c
henrygab: I removed the unused macro.  Later can define preferred debug output methods ... this was just a hack I needed for a short time.

src/ghostfat.c
hathach: thank you, I understand your concern, we don't expose or use this macro on pointer/function argument and is relatively rule out its pit-fall. We can come back to this when we need an global ARRAY_SIZE, or we could come back to this later on. 

PS: I take a quick look at simple hack, it is great hack. Though I think we could simplify its array size more, will do it in another PR later on.

src/ghostfat.c
hathach: thank you

hathach (commented):
Thank you @henrygab for actively helping with PRs. I really appreciate your review and inut, the ghostfat is getting better and better and could support pre-defined large file size. I think it gets to the point we could make an `Adafruit_GhostFAT` arduino library to help exposing data/log/variable (read-only) on board that have limited space or just don't want to deal with file system and/or partitioning.

hathach:
@henrygab let me know if you have any pending changes on your side, otherwise, I think we are good to merge this by now :tada: :tada:

henrygab:
I agree, this is good to go.  I was just waiting for the builds, and got distracted.  :)

Congratulations!  This will be a great, user-visible improvement!

PR #187, opened 2022-01-22
Improve FAT generation, bugs, favicon support

henrygab:
#### `.github/workflows/...`
* Moves self-test to new job runner (now ESP, ARM, self-test)

#### `ports/test_ghostfat/boards/...`
* updated known-good images (Note: This was done BEFORE updating how FAT was generated)
* add new boards to include a favorites icon
* slightly updated debug output

#### `src/ghostfat.c`
* Update macro `UF2_DIV_CEIL` -- avoid undefined behavior from overflow with signed inputs
* Change `TextContent` to `FileContent` to reflect ability to store binary files
* Fix latent bug in calculation of count of root directory sectors
* Cache starting cluster of each file (avoids O(n^2) runtime), use cached values for helper functions
* Simplify FAT generation, fix bugs with files whose fat chain crossed sector boundary

<details><summary>Click to expand FAT generation notes</summary><P/>

Previously, code presumed that (other than firmware file) the existing files would have FAT entirely in first sector.  This should now be fixed.

Generation of FAT uses much simpler method now, relying on the following properties:
1. All files are laid out contiguously, in order listed in table
2. No unused sectors between files
3. Last "file" is actually firmware (generated UF2)

New method of FAT generation is two steps:
1. Fill default value of CLUSTER+1  (or zero if past end of UF2) ... essentially make entire used FAT area a single, long, continuous FAT chain
2. Fill in the exceptions:
    a. clusters 0 & 1 need special values
    b. for each file, make end-of-FAT chain

That's it!  Simpler, and avoids the edge cases....
</details><HR/>


hathach (approved):
the new FAT generation method is brilliant !! Thank you very much for vastly improving the ghostfat and the self-test. There is other thing I think we could tweak the code, but I will do it as following commits.

PR #192, opened 2022-02-18
Add hiibot iots2

Neradoc:
Adding the Hiibot IOTS2.
PID from Espressif.

The IOTS2 board does not have a reset pin for the screen, it's pulled up.
So I set it to `-1` and added `#if DISPLAY_PIN_RST != -1` where appropriate so it compiles.
Don't know if it's the right way to do it.

### DIFF
--- a/ports/espressif/boards/hiibot_iots2/board.h
+++ b/ports/espressif/boards/hiibot_iots2/board.h
@@ -1,0 +1,95 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2020 Ha Thach (tinyusb.org) for Adafruit Industries
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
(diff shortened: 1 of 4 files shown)


ports/espressif/boards/hiibot_iots2/board.h
```diff
@@ -0,0 +1,95 @@
...
+// USB UF2
+//--------------------------------------------------------------------+
+
+#define USB_VID           0x80E7            // Espressif VID
+#define USB_PID           0x8111            // Espressif assigned PID
+#define USB_MANUFACTURER  "HIIBOT"
+#define USB_PRODUCT       "HiiBot IoTs2"
+
+#define UF2_PRODUCT_NAME  USB_MANUFACTURER " " USB_PRODUCT
+#define UF2_BOARD_ID      "HiiBoT-IoTs2"
```
hathach, line 93: board id must follow uf2 specs format https://github.com/microsoft/uf2#files-exposed-by-bootloaders

hathach (changes requested):
everything look great, the `DISPLAY_PIN_RST= -1` is good one. Only 1 request that BOARD_ID must follow uf2 format specs (check out other board for sample)

ports/espressif/boards/hiibot_iots2/board.h
Neradoc: Fixed ! Thanks.

hathach (approved):
look good, will merge when ci passed

makermelissa:
It doesn't appear this board was added to https://github.com/adafruit/tinyuf2/blob/master/.github/workflows/build_esp32.yml, so it wasn't part of the last release.

PR #197, opened 2022-03-28
[STM32F4] Allow user defined amount of Neopixels as status indicator

KarlK90:
...by defining NEOPIXEL_ NUMBER which now acts on values bigger than 1. Furthermore users can define a custom pin mode for the output pin, e.g. open-drain for applications where an pull-up resistor is used as a 3.3V to 5V logic level shifter.

This feature is handy for e.g. stm32f4 keyboards that have long external sk6812/ws2812 chains which can be used together with tinyuf2.

### DIFF
--- a/ports/stm32f4/boards.c
+++ b/ports/stm32f4/boards.c
@@ -71,7 +71,7 @@ void board_init(void)
 
 #if NEOPIXEL_NUMBER
   GPIO_InitStruct.Pin = NEOPIXEL_PIN;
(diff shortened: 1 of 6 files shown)


drashna:
If it helps, I can confirm that this works on a chain with ~20 leds, and with a pull-up resistor used.

KarlK90:
Great! Thanks for the feedback.

hathach (approved):
changes look great, thank you for the PR

PR #206, opened 2022-05-06
Fix esp32sx ci build

hathach:
- git throw `fatal: unsafe repository` when running git describe for getting version. Due to latest git update from ci ubuntu image 20.04. See https://github.com/actions/checkout/issues/766 for detail
- also explicitly specify `depth = 0`
- bump up other actions to latest version


zerocatchup (commented):
[**@**]()

PR #213, opened 2022-05-22
version with psram

ladyada:


hathach (approved):
`SPI_DMA_CH_AUTO` is the correct fix and a better usage of SPI channel. I have tested with 
- S2 TFT
- S3 devkit + TFT breakout
- Funhouse which use both SPI for TFT and DotStar and run fine

PR #215, opened 2022-06-11
stm32l4 support

m-mcgowan:
This PR implements a port of tinyuf2 to the STM32L4 family, specifically the L4R5. 

The code has been tested for a few months and is considered fairly stable. 

The implementation is mostly a copy and paste of the existing STM32F4 port, updated with the most recent changes.  In the longer run, the common elements of F4 and L4 ports may benefit from being merged to avoid code duplication. 

I tested the bootloader using this test plan on Swan R5:

* start with a fully erased device
* use dfu-util to load the bootloader binary and power-cycle the device
* verify the SWANBOOT drive appears
* copy a UF2 file to the drive (e.g. circuitpython)
* after the file has copied, verify that the CIRCUITPY drive appears
* double-tap RESET and verify the SWANBOOT drive appears. 
* single-tap RESET and verify the CIRCUITPY drive re-appears

### DIFF
--- a/ports/stm32l4/Makefile
+++ b/ports/stm32l4/Makefile
@@ -6,7 +6,7 @@ ST_CMSIS = lib/st/cmsis_device_l4
 CMSIS_5 = lib/CMSIS_5
 
 # List of git submodules that is included as part of the UF2 version
-GIT_SUBMODULES = st/cmsis_device_l4 st/stm32f4xx_hal_driver tinyusb
+GIT_SUBMODULES = st/cmsis_device_l4 st/stm32l4xx_hal_driver tinyusb
 
 include ../make.mk
 
--- a/ports/stm32l4/Makefile
+++ b/ports/stm32l4/Makefile
@@ -43,7 +43,7 @@ PORT_SRC_C += \
 
 SRC_C += \
 	$(PORT_SRC_C) \
-	lib/tinyusb/src/portable/st/synopsys/dcd_synopsys.c \
+	lib/tinyusb/src/portable/synopsys/dwc2/dcd_dwc2.c \
 	
 # Port include
 INC += \
--- a/ports/stm32l4/Makefile
+++ b/ports/stm32l4/Makefile
@@ -1,4 +1,4 @@
-UF2_FAMILY_ID = 0x57755a57
+UF2_FAMILY_ID = 0x00ff6919
 CROSS_COMPILE = arm-none-eabi-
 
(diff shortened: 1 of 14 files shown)


hathach (commented):
Thank you for your PR, I actually don't mind to have a bit of code duplication with f4, most stm32 families will have some shared code with each other. But they will have some difference especially flash API sooner or later. Also that make it easier to test/maintain, since the last thing we want with bootloader is unexpected behavior on L4 while doing some work with unrelated F4 mcu.

I could merge this PR as it is, but it would be nicer if you could
1. add your board to ci, to make sure we don't miss any files and/or having binaries of your board available in release https://github.com/adafruit/tinyuf2/blob/master/.github/workflows/build_arm.yml
2. Update Board-ID to follow uf2 spec.

m-mcgowan:
I'm not immediately sure why the build fails. The makefile works locally.

ports/stm32l4/Makefile
```diff
@@ -0,0 +1,62 @@
+UF2_FAMILY_ID = 0x00ff6919
+CROSS_COMPILE = arm-none-eabi-
+
+ST_HAL_DRIVER = lib/st/stm32l4xx_hal_driver
+ST_CMSIS = lib/st/cmsis_device_l4
+CMSIS_5 = lib/CMSIS_5
+
+# List of git submodules that is included as part of the UF2 version
+GIT_SUBMODULES = st/cmsis_device_l4 st/stm32f4xx_hal_driver tinyusb
```
hathach, line 9: st/stm32f4xx_hal_driver --> st/stm32l4xx_hal_driver

hathach (commented):
ci failed is due to typo in submodule

https://github.com/adafruit/tinyuf2/runs/6909281207?check_suite_focus=true#step:10:17

ports/stm32l4/Makefile
m-mcgowan: Thanks, I hadn't seen this :-)

m-mcgowan:
This one should be good to merge now :-)

hathach (approved):
thank you very much for the pr.

m-mcgowan:
My pleasure! Is there a recommended process to make the compiled STM32L4 bootloader binary available to developers?

hathach:
the compiled binaries is available for each ci commit https://github.com/adafruit/tinyuf2/actions/runs/2517588142

release binaries are also available under release page as well https://github.com/adafruit/tinyuf2/releases/tag/0.9.4 . Though you need to wait until our next release.

Issue #208, opened 2022-05-11
MIMXRT Port - Possible incorrect BOARD_BOOT_LENGTH  being stuffed into boot_data struct


ehughes:
### Operating System

Windows 10

### INFO_UF2.TXT

n/a

### What happened ?

**Summary:**

I believe the value use for BOARD_BOOT_LENGTH is incorrect.   In ports/mimxrt10xx/boards.h    

```
#define BOARD_BOOT_START        (((uint32_t) _ivt_origin) - 0x1000)

#define BOARD_BOOT_LENGTH       (0x8800) 
```

0x8800 is the value used for _text_length in ports/mimxrt10xx/linker/memory.ld

Since the ROM boot loader is copying down BOARD_BOOT_LENGTH from (((uint32_t) _ivt_origin) - 0x1000),   I believe  some code in the .text section will not get copied if you the image starts to get large. (only 0x8000 bytes)

I **think** the correct fix is something like

#define BOARD_BOOT_LENGTH   (0x8800 + 0x0800)

but it would probably be nice to have the values driven from the the values used in the .ld files.

**Backstory:**

The vanilla build of tinyuf2 for the MIMXRT1021 would yield a text section of 24792 bytes.  (0x60d8 bytes) which ran fine.

I noticed that as I added in the LOG=1 and a few extra init routines for an LCD on a board,  I would get goofy behavior on cold power up that would **not** occur when loading directly into itcm.    The CPU would end up hard faulting.   My modified program that enabled logging and some IO was getting close to the 0x8800 text section size (32560 bytes)

Changing BOARD_BOOT_LENGTH  to something like (0x8800 + 0x0800)  fixed the issue.    If I understand the linker .ld files correctly for the RT1021,   ivt + interrupts will use up an extra 0x0800 bytes that need to get copied down along with the text section

If you could please take a looking and verify my thinking/analysis.   I don't think anyone would notice this w/ vanilla builds but it come pop up on another RT board port if more code was added.

### How to reproduce ?

n/a

### Debug Log

n/a

hathach:
just checking, did we manage to fix it previously. I kind of forgot.

ehughes:
It has not been fixed.   I am pretty sure what I suggested is the fix.  BOARD_BOOT_LENGTH needs to be whatever _text_length (from the memory.ld file) + 0x2000 to work properly.   

I

hathach:
[...]

PR #220, opened 2022-06-20
Update boards.h

ehughes:
Patched boot length to include IVT structure and interrupt table.  Previous value would cause weird issues if program size got near the .text section length

See Bug #208


hathach (commented):
Superb !! I finally could wrap my head around this. This is indeed correct fix. Though I think we should best use all the linker symbol here to avoid the need to change this should we decide to expand the bootloader size in the future. I will try to make a update

hathach (approved):
I have updated to defined the `BOARD_BOOT_LENGTH` following the ideal suggestion in your previous comment. Which is based on the linker symbol. Tested with printf and the value is correctly 0x9000 now. Thank you very much for your PR.

PR #230, opened 2022-07-15
more rt10xx fix

hathach:
follow up to #220 

- fix write address of rt1020/rt1024 0x60000000
- add uart to rt1020 rt1024
- fix rt1024 LED
- update rt1024 ld ram to match 1020
- fix write_tinyuf2_to_flash() to also include fcbb as well.


ports/mimxrt10xx/board_flash.c
```diff
@@ -58,8 +61,11 @@ static void write_tinyuf2_to_flash(void)
   uint8_t const* image_data = (uint8_t const *) &qspiflash_config;
   uint32_t flash_addr = FCFB_START_ADDRESS;
 
+  // fcfb + bootloader (ivt, interrupt, text)
+  uint32_t const end_addr = FLEXSPI_FLASH_BASE + ((uint32_t) _fcfb_length) + BOARD_BOOT_LENGTH;
```
hathach, line 65: @ehughes following up to your PR, I think this also need changes as well. Let me know if this makes sense to you.

hathach: I will merge this as it is for now, should it has any issues, we could fix it in following PR. thank you.

PR #236, opened 2022-08-02
Add support for esp32-s3-eye devkit

jepler:
This depends on the PIDs that have been requested from espressif: 
 * https://github.com/espressif/usb-pids/pull/54


ports/espressif/build/CMakeFiles/cmake.check_cache
```diff
@@ -0,0 +1 @@
+# This file is generated by cmake for dependency checking of the CMakeCache.txt file
```
hathach, line 1: this file should not be added.

hathach (commented):
look great, you will need to add it into the ci build as well https://github.com/adafruit/tinyuf2/blob/master/.github/workflows/build_esp32.yml

ports/espressif/build/CMakeFiles/cmake.check_cache
jepler: I force-pushed to remove this file.

jepler:
I confirmed that the built firmware works for loading circuitpython uf2 files

jepler:
In theory support for the LCD could be added but it's not a priority for me

hathach:
> In theory support for the LCD could be added but it's not a priority for me

not a problem at all, we could do it later when having time. Please add it into the ci build and this is good to merge. https://github.com/adafruit/tinyuf2/blob/master/.github/workflows/build_esp32.yml

.github/workflows/build_esp32.yml
```diff
@@ -66,7 +66,9 @@ jobs:
         - 'adafruit_qtpy_esp32s3'
         - 'bpi_leaf_s3'
         - 'espressif_esp32s3_devkitc_1'
+        - 'espressif_esp32s3_devkitc_1'
```
hathach, line 69: This line is duplicated to above line of 68

jepler: oops, removed now!

hathach (approved):
perfect now

PR #239, opened 2022-08-14
rev b pins

ladyada:
small pin changes


hathach (approved):
look good to me. I try to find schematics in MBAdafruitBoards repo to confirm but couldn't find this board though.

PR #241, opened 2022-08-16
Add waveshare ESP32-S2-Pico-LCD

mayrhomar:
Adds support for the Waveshare ESP32-S2-Pico-LCD Board with USB PID from Espressif: [Espressif USB customer-allocated PID repository](https://github.com/espressif/usb-pids/blob/b39d288d4821bedfb589b7ec3cc8a3e2779e6237/allocated-pids.txt#L275)


hathach (approved):
look good, thank you

PR #244, opened 2022-09-12
Added Cytron Maker Feather AIoT S3.

waiweng83:
Added support for new board.

PR #245, opened 2022-09-13
Reduced the length for volume label.

waiweng83:
I'm sorry, just noticed that the length for volume label is > 11 character in #244.

By the way, GPIO 39 is high when entering UF2 bootloader. Is that normal?


hathach (approved):
look good

PR #250, opened 2022-10-11
Add BananaPi BPI-PicoW-S3 board.

Wind-stormger:
Support double reset to tinyUF2.
PID from Espressif.

### DIFF
--- a/ports/espressif/boards/bpi_picow_s3/board.h
+++ b/ports/espressif/boards/bpi_picow_s3/board.h
@@ -1,0 +1,68 @@
+/* 
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2020 Ha Thach (tinyusb.org) for Adafruit Industries
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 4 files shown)


ports/espressif/boards/bpi_picow_s3/board.h
```diff
@@ -0,0 +1,68 @@
...
+// USB UF2
+//--------------------------------------------------------------------+
+
+#define USB_VID                  0x303A
+#define USB_PID                  0x812D
+#define USB_MANUFACTURER         "BananaPi"
+#define USB_PRODUCT              "BPI-PicoW-S3"
+
+#define UF2_PRODUCT_NAME         USB_MANUFACTURER " " USB_PRODUCT
+#define UF2_BOARD_ID             "BPI-PicoW-S3"
```
hathach, line 64: board id need to follow uf2 specs https://github.com/microsoft/uf2#files-exposed-by-bootloaders

hathach (changes requested):

ports/espressif/boards/bpi_picow_s3/board.h
Wind-stormger: Thanks for the review, fixed.

hathach (changes requested):
one more thing, you need to add this board to ci so that its binary is available in the release https://github.com/adafruit/tinyuf2/blob/master/.github/workflows/build_esp32.yml

Wind-stormger:
Thanks for the reminder, added.

hathach (approved):
look good now

PR #251, opened 2022-10-13
Add Espressif ESP32-S3 BOX

Neradoc:
- PID from Espressif: https://github.com/espressif/usb-pids/blob/main/allocated-pids-espressif-devboards.txt#L12
- Link goes to the github repo: https://github.com/espressif/esp-box
- Checked that the side "boot mode" button can get into UF2 bootloader
- Successfully installed Circuitpython with it

This does not include the `lite` version, don't know if it can just use the same build.


hathach (approved):
thank you, I think the lite version is pretty the same and can actual use this binary right ?

PR #275, opened 2023-01-18
Align stm32f3 port with stm32f4

zvecr:
## Description of Change

While implementing self update and write protection for stm32f3, I figured it was worth a separate PR for the general tidy up. This should help make later PRs easier to digest, and shouldn't dramatically change the build output in the short term.


hathach (approved):
Superb !! Thank you very much for your PR. F3 indeed is lacking behind.

Issue #277, opened 2023-01-31
Update GitHub Actions to Upload to Amazon S3


makermelissa:
In order to make the bootloader accessible for ESP32 Devices to the New CircuitPython Installer I'm working on and get around CORS issues, GitHub Actions needs to be updated to upload the release files to Amazon S3 like we do with CircuitPython Libraries and Firmware currently.

This should be fairly easy as we already have it implemented in other repos.

hathach:
implemented by #278

PR #278, opened 2023-02-01
Update ESP32 GitHub Actions to upload to Amazon S3

makermelissa:
-----------

## Description of Change

This modifies the GitHub actions so that release files are uploaded to Amazon S3. I was unable to test this in my own repo. I already set up the AWS secrets, so this should work. I'll do a release to test after this is merged.

If the test works, then #277 should be done.


hathach (approved):
look great

PR #279, opened 2023-02-02
Fix pathname for AWS uploads

makermelissa:
## Description of Change

After testing #278, I realized it was uploading everything to a file called `esp32` instead of using that as a folder. This fixes that issue. Otherwise, the rest of the testing worked perfectly.


makermelissa:
Since I'm 100% sure this will work and it needs to be changed ASAP, I'm just going to merge.

PR #280, opened 2023-02-02
Add release tag to pathname for AWS uploads

makermelissa:
## Description of Change

I realized while #279 did indeed fix the issue with it using the filename instead of the folder name, it was missing the release tag. This fixes that issue and I'd be surprised if another follow-up PR is needed.


makermelissa:
Ok, it worked perfect this time: https://adafruit-circuit-python.s3.amazonaws.com/index.html?prefix=bootloaders/esp32/

PR #282, opened 2023-02-07
Update for metro m7

hathach:
## Description of Change

This PR includes update for new Metro M7 rt1011 and bunch of other update as well including
- update flashconfig for metro m7 to match new flash device W25Q64JVSIQ
- add option for imxrt tinyuf2 to compare its running image at ram and flash contents and write to flash if not matched. Though leaving as disabled for now.
- add explicit `get-deps` target to fetch submodule dependencies
- fix missing library name in UF2_INFO.TXT in some cases
- fix esp32programmer with uart read issue, also implement dtr/rts signal manipulation (but leaving disabled to be compatible with `--before no_reset` option
- update tinyusb to commit https://github.com/hathach/tinyusb/commit/c0d79457f61cc4ee27336f430a6f96403bd8b289
- add factory_test app for metro tester
- add optional (for now) board_teardown() api
- add aarch64 binary for sdphost since it is rather complicated to install pypip spsdk in rpi os (require rust to compile cmsis-package manager)


ports/mimxrt10xx/apps/esp32programmer/main.c
```diff
@@ -38,95 +38,151 @@
  * used to program ESP32 Co-Processors
  */
 
+// Enable this for more reliable connection require esptool.py default reset option "--before default_reset"
+// i.e "--before no_reset" should not be include in the esptool.py command
+#define ESP32_DTR_RTS_BOOT_RESET_SUPPORT    0
```
hathach, line 43: DTR/RTR signal manipulation with esptool.py is supported but not enabled to be compatible with `--before no_reset` option used in adafruit learn guide/tutorial. Can be enabled if needed.

PR #288, opened 2023-02-25
STM32F3: Implement self-update

zvecr:
## Description of Change

Mostly alignment to STM32F4's self-update implementation.

The bulk of the changes align `board_self_update` and `flash_write`, as well as duplicating all the ancillary files.


ports/stm32f3/board_flash.c
```diff
@@ -71,59 +71,76 @@ static bool is_blank(uint32_t addr, uint32_t size)
...
-
-    TUF2_LOG1("Erase: %08lX size = %lu\n", addr, size);
+    TUF2_LOG1("Erase: %08lX size = %lu KB ... ", sector_addr, size / 1024);
 
     FLASH_EraseInitTypeDef EraseInit;
     EraseInit.TypeErase = FLASH_TYPEERASE_PAGES;
-    EraseInit.PageAddress = addr;
-    EraseInit.NbPages = ((0x08040000 - addr)/size);
+    EraseInit.PageAddress = sector_addr;
+    EraseInit.NbPages = ((BOARD_FLASH_APP_START - sector_addr)/size);
```
hathach, line 112: I know this is an historical value, but looks like `EraseInit.NbPages` should always be 1. sector_addr can be larger than the `BOARD_FLASH_APP_START` when programming an actual app, I have no idea why I put this value in the first place. Would you mind doing a quick check to comfirm this.

hathach (commented):
Thank you for the PR, this looks perfect. Though I have an favor to ask, would you mind checking the erase nbpages since it looks awkward to me even though it exists previously, we should fix it if needed.

ports/stm32f3/board_flash.c
zvecr: Number of pages would be 8 on the default start address (16K of 2K pages). However I have code locally that adds additional board configuration, which currently goes over that 16K marker and would want to keep this area somewhat dynamic so that its less areas to update.

hathach: the flash_erase() is called per each sector, so I think this should just be 1, yes ?

zvecr: I want to say `flash_write` is called with chunks larger than a page, though I will have to give this a proper look when im back at a machine.

hathach: look good, let me make the changes.

ports/stm32f3/board_flash.c
```diff
@@ -71,59 +71,76 @@ static bool is_blank(uint32_t addr, uint32_t size)
...
+    uint32_t SectorError = 0;
     HAL_FLASHEx_Erase(&EraseInit, &SectorError);
     FLASH_WaitForLastOperation(HAL_MAX_DELAY);
-
-    if (SectorError != 0xFFFFFFFF)
-    {
-      TUF2_LOG1("failed to erase!\r\n");
-    }
+    TUF2_LOG1("OK\r\n");
+    TUF2_ASSERT( (SectorError != 0xFFFFFFFF) && is_blank(sector_addr, size) );
```
hathach: seems to be inverted logic error, it must be `TUF2_ASSERT(SectorError == 0xFFFFFFFF)` since 0xFFFFFFFF all pages is erased.

hathach (approved):
I made changes according to our discussion, also fix an assert check with SectorError (see review). Let me know if these changes still works for you, if yes, then we could merge this PR

ports/stm32f3/board_flash.c
zvecr: As long as there is the assumption that `flash_write` does not get called with an address+length that would cross 2 flash pages, I think the code should be okay.

hathach: for now, it won't do that, so I guess we are good (for now)

hathach:
> LGTM, not found any additional issues with my testing.

thank you for testing, merged !!!

PR #289, opened 2023-02-25
Refactor STM32 flashers

zvecr:
## Description of Change

* Support erase for `dfu-util`
  * Moved to common location to avoid duplication
* Allow use of `st-flash` CLI as an alternative to `STM32_Programmer_CLI`


hathach (changes requested):
Look great, though I think we should add an additional `flash-stflash` and `erase-stflash` instead of using the current stlink. What do you think ?

zvecr:
I did have that previously, and its use can be hidden with a `:flash` alias. Though I do think its somewhat unintuitive for  `flash-stlink` to not work with the connected stlink.

hathach (approved):
look good, thank you

PR #301, opened 2023-03-07
Allow per-board custom init, teardown, and validate hooks

ccrome:
## Checklist
- [x] Please provide related links (*eg. Issue which will be closed by this Pull Request*)

-----------

## Description of Change
Some custom i.MX boards require per board setup.  Currently, I don't see a mechanism for a specific board to have a setup call.

As a specific example, For example, on the i.MX RT 1011, if you happen to have the boot LED on `GPIO_AD_02`, you might need to custom configure `GPR26` in order for it to work with the GPIO configuration you want to use.   

Additionally, there are lots of other potential reasons you might want to do some setup in `boards/<board_name>`.

This patch adds `#ifdef BOARD_INIT_POST_HOOK `, and if it's defined, then `board_init_post_hook()` will be called at the end of `board_init()`.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -63,6 +63,7 @@ static bool check_dfu_mode(void);
 int main(void)
 {
   board_init();
+  if (board_init2) board_init2();
   TU_LOG1("TinyUF2\r\n");
 
 #if TINYUF2_PROTECT_BOOTLOADER
@@ -74,6 +75,7 @@ int main(void)
   {
     TU_LOG1("Jump to application\r\n");
     if (board_teardown) board_teardown();
+    if (board_teardown2) board_teardown2();
     board_app_jump();
     while(1) {}
   }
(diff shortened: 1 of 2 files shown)


ccrome:
It does beg the question, are there other places where there should be custom hooks on a per-board basis?

hathach:
we add an optional API (weak function) to board_api.h e.g `board_init_post()` in conjunction with board_teardown_post()

ccrome:
I updated the hooks to be in main now, rather than in the port directory.  Now it's easy for any board anywhere to define the functions to do per board custom stuff.

```
void board_init_post(void) __attribute__ ((weak));
void board_teardown_post(void) __attribute__ ((weak));
bool board_app_valid_post() __attribute__ ((weak));
```

src/main.c
```diff
@@ -121,7 +123,10 @@ static bool check_dfu_mode(void)
 #endif
 
   // Check if app is valid
-  if ( !board_app_valid() ) return true;
+  bool valid = board_app_valid();
```
hathach, line 126: should be
```C
if ( !board_app_valid() ) return true;
if ( board_app_valid_post && !board_app_valid_post() ) return true;
```

hathach (changes requested):
Please follow review for suggested change. Also I changed my mind on the naming, please change these _post to simply
- board_init2()
- board_teardown2()
- board_app_valid2()

ccrome:
I think I've got it all.  Thanks!

hathach (approved):
all good, thank you for your PR

PR #302, opened 2023-03-09
imxrt10xx: fix LED blinking

ccrome:
## Checklist

-----------

## Description of Change
The LED write routine is meant to take a value between 0 for off and 255 for full brightness. the current 10xx port treats 0 as off, and anything else as on when dealing with non-PWM leds.  This creates a case where the LED is on virtually all the time, except for a tiny blink off every 256 timer intervals. 
This fix turns the LED on for `(value >= 128)`  and off for (value < 128), and makes the LED blinking look correct.


hathach (approved):
look good, thank you

PR #309, opened 2023-04-04
pull request for new esp32s3 dev board 

CircuitART:
link to VID/PID 
https://github.com/espressif/usb-pids

USB_VID = 0x303A - USB_PID = 0x80DC | CircuitArt FeatherS3 - UF2 Bootloader

board has been added to /.github/workflows 

UF2 board id info
#define UF2_BOARD_ID             "ESP32S3-Zero-R1"
#define UF2_VOLUME_LABEL         "ZEROS3BOOT"
#define UF2_INDEX_URL            "https://github.com/CircuitART"

### DIFF
--- a/ports/espressif/boards/circuitart_zero_s3/board.h
+++ b/ports/espressif/boards/circuitart_zero_s3/board.h
@@ -1,0 +1,109 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2020 Ha Thach (tinyusb.org) for Adafruit Industries
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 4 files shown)


ports/espressif/boards/circuitart_zero_s3/board.h
```diff
@@ -0,0 +1,109 @@
...
+#define USB_MANUFACTURER         "CircuiArt"
+#define USB_PRODUCT              "ZeroS3"
+
+#define UF2_PRODUCT_NAME         USB_MANUFACTURER " " USB_PRODUCT
+#define UF2_BOARD_ID             "ESP32S3-Zero-R1"
+#define UF2_VOLUME_LABEL         "ZEROS3BOOT"
+#define UF2_INDEX_URL            "https://github.com/CircuitART"
+
+// Use favicon
+#define TINYUF2_FAVICON_HEADER   "favicon_adafruit_256.h"
```
hathach, line 107: you probably forgot to remove this line. The favicon will show up Adafruit logo when using with windows.

CircuitART: hi, oh okay got it will remove the FAVICON, but i also did some other changes as well which I will edit later.

by the way, as you might have noticed, my board has a built-in TFT (with no Pin for BL control) and i was wondering what I can type instead of the pin number for this
 #define DISPLAY_PIN_BL        46
i have tried -1 but it does not work because MISO has -1
i also used  NULL/null but i get errors during the build process

thanks

hathach: oh, you will need to slightly update the code for the lcd component to opt out when DISPLAY_PIN_BL = -1 https://github.com/adafruit/tinyuf2/blob/master/ports/espressif/components/lcd/lcd.c

CircuitART: i had a look at the lcd.c and i think its not within my knowledge so, for now, I used it as is.
thanks

hathach (approved):
look good now

PR #317, opened 2023-04-28
Add iMX RT 1015, 1040 and 1050 EVK support

tannewt:
Also includes changes for 1176 but I couldn't get it loading so I omitted the board def. It doesn't support sdphost. It'll need blhost support instead.


hathach (approved):
all changes is great, also tested on metro m7

PR #323, opened 2023-06-05
Minor alignment of STM32L4 port

zvecr:
## Description of Change

STM32L4 support has been mentioned a few times on the QMK Discord. Eventually targeting single-bank MCU like the STM32L432, there are a few assumptions on flash layout to unpick. 

Before submitting those changes, I figured it was worth a pass to tidy a few areas up so that its more consistent with the F3 and F4 port.


hathach (approved):
look great, thank you for updating L4 ports

PR #330, opened 2023-06-28
add qtpy s3 n4r2 (4M flash, 2 psram)

hathach:
## Description of Change

Add qtpy s3 n4r2 (4M flash, 2 psram). @ladyada I am not sure if the n4r2 naming look good to you, please let me know if you have any other suggestion.


ladyada (approved):
Tested and works!

PR #341, opened 2023-07-26
adafruit-camera-esp32s3: make LCD work

jepler:
## Description of Change

There were several bits of code that needed to be revised so that the LCD display works on the adafruit-camera-esp32s3:
 * the backlight has to be turned on through the I/O expander, which is an AW9523 type
 * the VSCSAD register value was wrong, making the top part of the screen corrupt

This is tested on PyCam Rev C board.


hathach (approved):
Look great, thank you

jepler:
per ladyada, a future rev of the camera board will ensure the LCD backlight is on without talking to the I/O expander so we may be able to back that code out in the future.

hathach:
No problem, we can revert it when the hw is out

PR #344, opened 2023-08-17
SUKU run the GPIO detection at least once even if UF2_DETECTION_DELAY…

SukuWc:
## Description of Change
- UF2_DETECTION_DELAY_MS is useful to allow ESP32 boards that only have a button on BOOT0 pin to use both ROM bootloader and UF2 bootloader.
- If however the board uses separate gpio for UF2 detect button, then this just slows down the boot procedure. I this case I wanted to set UF2_DETECTION_DELAY_MS to zero so that uf2 bootloader is only activated if the button was pressed before power on!
- In the gpio detection logic the while statement never evaluated true because more than 0 ms has already passed, making it impossible to enter bootloader mode!
- Do ... while syntax allows the detection logic to run at least once regardless of the UF2_DETECTION_DELAY_MS value!


hathach (approved):
look good, thank you

PR #351, opened 2023-10-29
Add Waveshare ESP32 S3 Pico

tirtharajsinha:
Request to add support for Waveshare ESP32 S3 Pico.
Circuitpython Support [added](https://github.com/adafruit/circuitpython/pull/8352).
VID/PID is provided by [Espressif](https://github.com/espressif/usb-pids/blob/main/allocated-pids.txt) 

```
0x81A2 | Waveshare ESP32-S3-Pico - UF2 Bootloader
0x81A3 | Waveshare ESP32-S3-Pico - CircuitPython
```

### DIFF
--- a/.github/workflows/build_esp32.yml
+++ b/.github/workflows/build_esp32.yml
@@ -9,6 +9,7 @@ on:
       - created
 
 jobs:
+
   # ---------------------------------------
   # Build ESP32SX family
   # ---------------------------------------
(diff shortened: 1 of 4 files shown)


tirtharajsinha (commented):
Why the arm board's workflow failed when there is no changed files that are arm related. I'm really confused right now.

hathach (approved):
look good, thank you. CI failed due to unrelated issue of python package (fixed in another PR).

PR #363, opened 2023-11-21
Espressif 16MB flash memory layout with 4MB app/ota partitions

lyusupov:
## Checklist

-----------

## Description of Change

There are applications for ESP32-S3 that soon will no longer fit 2MB APP/OTA partition size.
This project is an example of the application: https://github.com/nakhonthai/ESP32APRS_T-TWR

I've created this new layout for 16MB flash chips to support development of such application(s) and use it together with the TinyUF2 bootloader.


hathach (approved):
look good, note that uf2 address is changed to 0x810000 with this partition

PR #372, opened 2023-12-28
Only show purple waiting color on led if UF2_DETECTION_DELAY_MS is not zero

SukuWc:
## Description of Change

UF2_DETECTION_DELAY_MS can be set to zero when PIN_BUTTON_UF2 is not GPIO0 on ESP32. This causes the board to boot application firmware instantly if the PIN_BUTTON_UF2 is not pulled low by pressing the boot button. The problem is that the RGB LEDs still turn purple momentarily indicating that the bootloader is waiting for button press. This causes unwanted flash of the LEDs.  In this PR code skips board_led_on() is only called if UF2_DETECTION_DELAY_MS>0 condition is met!

Please note that for compatibility reasons if PIN_DOUBLE_RESET_RC is defined then board_led_on() will always be called regardless of UF2_DETECTION_DELAY_MS!


hathach (approved):
thank you, changes look good. I take the chance to also push some update to the code while testing it out as well.

PR #377, opened 2024-02-26
Add partions 4mb with 3.5mb app partition

hathach:
## Description of Change

Add partions 4mb with 3.5mb app partitionn


hathach:
note required anymore, but still handy to have

PR #381, opened 2024-04-10
Add lolin s3 mini

skieast:
## Description of Change

Add support for Lolin S2-Mini board
PID was already allocated in Espressif USB-PIDS repository
0x8169 | LOLIN S3 Mini - UF2 Bootloader


hathach (approved):
look great, thank you

PR #389, opened 2024-06-04
Initial support for ch32v20x (not fully working yet)

hathach:
## Description of Change

Initial support for ch32v20x, build and run, able to enumerate as MSC and written to flash. Double tap dectection work. However, @ladyada it is not fully functional since jump to app not working yet (hopefully other can help). But we made an good progress, will wrap this up for now, since it already takes quite a few days, will come back after doing other pending work (pr/issue).

Ch32v20x current state:
- Able to do double tap detection using 16-bit backup register R10
- usb msc enumerate with USBD (fsdev) and able to erase/program using fast page erase/program (a bit slow, will switch to caching and 4K erase later).
- led/timer/uart work well
- jumping to app does not work yet, need a bit more dive to riscv (need help, or work on this as follow up PR later)
- bootloader size is relatively large, no debug is 13KB, current set to 24KB for debug log. Though it should best to be under 8K since the chip is only 64KB in flash. Will do size optimization later on with follow-up PR.

```
   text	   data	    bss	    dec	    hex	filename
  13660	    724	   9080	  23464	   5ba8	_build/ch32v203_r0_1v0/tinyuf2-ch32v203_r0_1v0.elf
```

Lots of other Tinyuf2 refactor
- rename TINYUF2_DFU_DOUBLE_TAP to TINYUF2_DBL_TAP_DFU, DBL_TAP_REG to TINYUF2_DBL_TAP_REG, also add TINYUF2_DBL_TAP_REG_SIZE since ch32 backup register is only 16 bit. Make TINYUF2_DBL_TAP_DELAY configurable per port/board
- Change tusb_config.h to drop CFG_TUSB_RHPORT0_MODE, use tud_init() instead of tusb_init()
- change board_flash_write() return from void to bool
- prefer to use RAM intead of ROM for literal/const such as configuration descriptor
- more ci improvement for ghostfat testing


ladyada:
ok thanks. it is a little unclear about flash memory. some folks say theres 224K on a variant

https://cpldcpu.com/2024/05/01/decapsulating-the-ch32v203-reveals-a-separate-flash-die/

hathach:
oh, that is an interesting read, the official doc only mentnion max 64KB for v20x. I will give it a try to see if that is possible :) 

image

PR #391, opened 2024-06-05
Added UM FeatherS3 Neo.

UnexpectedMaker:
Added support for my new FeatherS3 Neo board.

PID/VID supplied by Espressif


hathach (approved):
look good

PR #404, opened 2024-08-15
Added 3 new Unexpected Maker boards

UnexpectedMaker:
Added my new OMGS3
Added my new RGB Touch Mini
Added my TinyWATCH S3

TinyWATCH S3 doesn't have a status LED or RGB LED, so I commented out the DATA and POWER pins, and set the number of RGB LEDs to 0.

I hope that is the correct way to handle this use case. It seems to work ok, but I am not sure if there's an alternate way to handle no status LED.

PIDs/VID supplied by Espressif


hathach (approved):
thanks, that is ok, when there is no NEOPIXEL_PIN defined, the rest won't be used.

UnexpectedMaker:
Thanks :)

PR #405, opened 2024-08-29
Revision 2 of CircuitART Zero S3

CircuitART:
This is Revision 2 of a board that has already been added through an older pull request by me.


hathach (approved):
look good, thank you

PR #414, opened 2024-10-08
ports/mimxrt10xx/boards: Add board files for Seeed ARch Mix, Olimex RT1011 and Makerdiary RT1011 Nano Kit..

robert-hh:
Boards:

- Seed Arch Mix RT1052
- Olimex RT1011
- Makerdiary RT1011 Nano Kit

All sets differ only slightly from the respective imxrt10xx-evk files. Major changes are flash size and LED pin. All have been tested with theses boards.

### DIFF
--- a/ports/mimxrt10xx/boards/makerdiary_rt1011/board.h
+++ b/ports/mimxrt10xx/boards/makerdiary_rt1011/board.h
@@ -1,0 +1,99 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2020 Ha Thach for Adafruit Industries
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ * This file is part of the TinyUSB stack.
+ */
+
+
+#ifndef BOARD_H_
+#define BOARD_H_
+
+// Size of on-board external flash
+#define BOARD_FLASH_SIZE     (16*1024*1024)
+
+//--------------------------------------------------------------------+
+// LED
+//--------------------------------------------------------------------+
+
+#define LED_PINMUX            IOMUXC_GPIO_SD_04_GPIO2_IO04
+#define LED_PORT              GPIO2
+#define LED_PIN               4
+#define LED_STATE_ON          1
+
+#if 0
+// PWM Test on Arduino header D8
+#define LED_PWM_PINMUX        IOMUXC_GPIO_SD_02_FLEXPWM1_PWM0_A
+#define LED_PWM_BASE          PWM1
+#define LED_PWM_MODULE        kPWM_Module_0
+#define LED_PWM_CHANNEL       kPWM_PwmA
+#endif
+
+//--------------------------------------------------------------------+
+// Neopixel
+//--------------------------------------------------------------------+
+
+// Number of neopixels
+#if 1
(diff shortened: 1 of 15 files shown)


ports/mimxrt10xx/boards/makerdiary_rt1011/board.h
```diff
@@ -0,0 +1,99 @@
...
+// USB UF2
+//--------------------------------------------------------------------+
+
+#define USB_VID           0x2886
+#define USB_PID           0xf00f
+#define USB_MANUFACTURER  "Makerdiary"
+#define USB_PRODUCT       "iMX RT1011 Nano Kit"
+
+#define UF2_PRODUCT_NAME  USB_MANUFACTURER " " USB_PRODUCT
+#define UF2_BOARD_ID      "iMX RT1011 Nano Kit"
```
hathach, line 87: should follow the board-id format https://github.com/microsoft/uf2?tab=readme-ov-file#files-exposed-by-bootloaders

```
UF2 Bootloader v1.1.3 SFA
Model: Arduino Zero
Board-ID: SAMD21G18A-Zero-v0
```

ports/mimxrt10xx/boards/olimex_rt1010/board.h
```diff
@@ -0,0 +1,99 @@
...
+// USB UF2
+//--------------------------------------------------------------------+
+
+#define USB_VID           0x15ba
+#define USB_PID           0x0046
+#define USB_MANUFACTURER  "Olimex"
+#define USB_PRODUCT       "RT1010"
+
+#define UF2_PRODUCT_NAME  USB_MANUFACTURER " " USB_PRODUCT
+#define UF2_BOARD_ID      "Olimex-RT1010"
```
hathach, line 87: see above board id format

hathach (commented):
Thank you for your PR, everything look goods, there is a minor request to change board-id to match uf2 specs format.

Also I have a question regarding VID/PID, are these assigned specifically for these board for bootloader purpose and re-use/shared with other application/board.

ports/mimxrt10xx/boards/makerdiary_rt1011/board.h
robert-hh: Thank you for the hint about the board ID structure. I'll change them accordingly.

robert-hh:
The VID/PID are set on best knowdledge. 
Olimex RT1011: PID and VID have been supplied by the vendor and are set by the APP as well.
Makerdiary RT1010 Nano Kit: The VID/PID are the values shown in their bootstrap loader. So I use the same here, even if the VID is one of Seeed Technologies.
Seeed Arch Mix:  0x2886 is tve VID of Seeed. No specific PID was supplied by the vendor. But I can ask.

hathach:
> The VID/PID are set on best knowdledge. Olimex RT1011: PID and VID have been supplied by the vendor and are set by the APP as well. Makerdiary RT1010 Nano Kit: The VID/PID are the values shown in their bootstrap loader. So I use the same here, even if the VID is one of Seeed Technologies. Seeed Arch Mix: 0x2886 is tve VID of Seeed. No specific PID was supplied by the vendor. But I can ask.

I guess olimex and makerdiary is OK, for Seed Arch, maybe if you could just ask to see if they allocated any PID for the board (you can @ their developer here know one).

gsteiert:
I requested NXP set aside a VID/PID that could be used for generically for
UF2. I would suggest using that for UF2 on NXP based boards. Otherwise you
always need two PID if your application has a different USB configuration
than UF2.

Greg

robert-hh:
@hathach Seeed responded and told me, that they cannot tell me a PID for the Arch Mix board. Maybe they never assigned one, or they could not find the number. The board I have is not any more in the delivery state, so I cannot just check. Anyhow I've set it to 0x0010. Maybe 0x1052 would be a good number as well.

hathach:
> I requested NXP set aside a VID/PID that could be used for generically for UF2. I would suggest using that for UF2 on NXP based boards. Otherwise you always need two PID if your application has a different USB configuration than UF2. Greg

That sounds great, do you think that makes sense to ask for a PID per mcu since they are a bit different from each other. Though I think a general one for imxrt is OK as well for board that does not has a private PID. Btw, @gsteiert since you are here, do you happen to know if there is any issue with imxrt1015, robert only get it working when compiled with gcc v10 https://github.com/adafruit/tinyuf2/discussions/413#discussioncomment-10853224 . I forgot the detail, but could look at this later if you don't have time.

> @hathach Seeed responded and told me, that they cannot tell me a PID for the Arch Mix board. Maybe they never assigned one, or they could not find the number. The board I have is not any more in the delivery state, so I cannot just check. Anyhow I've set it to 0x0010. Maybe 0x1052 would be a good number as well.

The rule of thumb is making sure that the VID/PID is valid and unique if possible (not conflicting with other existing device whose vid/pid is owned by other). If your PID for seed it not used elsewhere then I think we are good with since I think seed agree for you to use their VID. Otherwise we can also wait for greg (FAE from NXP) to see if he could get an generic PID from NXP.

ports/mimxrt10xx/port.mk
```diff
@@ -1,6 +1,7 @@
 UF2_FAMILY_ID = 0x4fb2d5bd
 CROSS_COMPILE = arm-none-eabi-
 
+# SDK_DIR = /home/robert/Downloads/mcux-sdk
```
hathach, line 4: probably need to remove this

hathach (commented):
there seems a typo in port.mk

ports/mimxrt10xx/port.mk
robert-hh: Indeed, it is. I'II remove it.

robert-hh: Updated.

hathach (approved):
perfect, thank you

hathach:
you're welcome, I am still stuck with dwc2 host driver for the time being. Will test out the imxrt1015 later on when possible

PR #420, opened 2024-11-11
ports/stm32f4: Keep bootloader within first 32K

petejohanson:
In order to free up sectors 2 and 3 for the application, limit the bootloader to the first 16K sector, config to the next 16K. We keep the application still starting at 0x10000 so that 0x8000 and 0xC000 sectors can be used by the application for settings/config storage.

## Checklist

*By completing this PR sufficiently, you help us to review this Pull Request quicker and also help improve the quality of Release Notes*

- [ ] Please provide related links (*eg. Issue which will be closed by this Pull Request*)

*This checklist items that are not applicable to your PR can be deleted.*

-----------

## Description of Change

While trying to set up an stm32f401cb based board to run ZMK (using Zephyr under the hood), I ran into an issue trying to use Zephy's NVS driver for persistence of settings. Zephyr's NVS expects two sectors of 16K, which basically means I need to use sectors 2 and 3 for this:

image

I was able to make this work with the correct partitioning on the Zephyr/ZMK side, and properly fit the bootloader into the first two sectors with the changes in this PR. The device runs, and properly store settings as expected.

### DIFF
--- a/ports/stm32f4/linker/stm32f4_boot_compact.ld
+++ b/ports/stm32f4/linker/stm32f4_boot_compact.ld
@@ -1,0 +1,176 @@
+/*
+*****************************************************************************
+**
+
+**  File        : LinkerScript.ld
+**
+**  Abstract    : Linker script for STM32F405RGTx Device with
+**                1024KByte FLASH, 128KByte RAM
+**
+**                Set heap size, stack size and stack location according
+**                to application requirements.
+**
+**                Set memory bank area and size if external memory is used.
+**
+**  Target      : STMicroelectronics STM32
+**
+**
+**  Distribution: The file is distributed as is, without any warranty
+**                of any kind.
+**
+**  (c)Copyright Ac6.
+**  You may use this file as-is or modify it according to the needs of your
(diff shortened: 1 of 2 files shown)


petejohanson:
So, based on the CI failure... A few things are clear:

* tinyuf2 only fits with bare minimum features enabled. Feather build fails, for example, when the neopixel code gets built too. Setting the number of pixels to 0 for that board makes the build pass again here locally.
* I realize that changing this also means the config location would change for a lot of boards, probably breaking things for existing devices.

Thoughts on making this an optional different linker file/set of offsets that boards can opt in to?

petejohanson:
Ok, I've made this change "opt in" by setting `COMPACT_BOOTLOADER` in `board.mk`, so existing boards don't have any change to the flash usage. Thoughts?

kbladewht:
Thanks your reply, yes, I am not sure where is the issue, after check log , it stuck on  __set_MSP(sp); from board.c

image

petejohanson:
Sorry, I clearly needed more coffee/sleep when working on this. This keeps it in the first 16K, but that's really not needed for my use case, so closing.

PR #424, opened 2024-12-31
Adding support for LILYGO T-Display S3

rgrizzell:
VID/PID allocation: https://github.com/espressif/usb-pids/blob/388e9af066ac0b88342a397c208b0a6c05043a2e/allocated-pids.txt#L328

## Display

The display uses a Parallel Bus instead of SPI, so I was not able to get that working with this build. However, I defined the pins and can re-test if support for Parallel Bus is added in the future.

## Testing

Build was compiled and flashed using ESP-IDF v5.1.4

Output from `INFO_UF2.TXT`
```
TinyUF2 Bootloader 0.20.1-25-g46de518 - tinyusb (0.16.0-1090-g3eea46056)
Model: LILYGO T-Display S3
Board-ID: ESP32S3-T-Display-S3
Date: Dec 30 2024
Flash Size: 0x00200000 bytes
```

The `INDEX.HTM` file working as intended.
```
<!doctype html>
<html><body><script>
location.replace("https://github.com/Xinyuan-LilyGO/T-Display-S3");
</script></body></html>
```

I then copied the [CircuitPython 9.2.1 UF2 build for the T-Display S3](https://downloads.circuitpython.org/bin/lilygo_tdisplay_s3/en_US/adafruit-circuitpython-lilygo_tdisplay_s3-en_US-9.2.1.uf2) to the test device and it started up.
```
Adafruit CircuitPython 9.2.1 on 2024-11-20; LILYGO T-DISPLAY S3 v1.2 with ESP32S3
Board ID:lilygo_tdisplay_s3
```

Pressing the `RESET` button and then `BUTTON0` puts it back into the UF2 bootloader.

### DIFF
--- a/ports/espressif/boards/lilygo_tdisplay_s3/board.cmake
+++ b/ports/espressif/boards/lilygo_tdisplay_s3/board.cmake
@@ -1,0 +1,2 @@
+# Apply board specific content here
+set(IDF_TARGET "esp32s3")
--- a/ports/espressif/boards/lilygo_tdisplay_s3/board.h
+++ b/ports/espressif/boards/lilygo_tdisplay_s3/board.h
@@ -1,0 +1,105 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2024 Robert Grizzell, Independently providing these changes
+ *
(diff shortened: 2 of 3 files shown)


hathach (approved):
look good, thank you

PR #425, opened 2025-01-11
Waveshare pico S3: change the color order of the RBG LED

Neradoc:
I didn't find a clear documentation of the color order of the neopixel on the board, but the Micropython demo from Waveshare does swap red and green instead of the default (1, 0, 2, 3), so I think it's good.
```py
# Set the color order for the NeoPixel object
rgb_led.ORDER = (0, 1, 2, 3)
```
Tested and the color is now correct (LED green when connected to USB, red when on a power bank).


hathach (approved):
look good, thank you

PR #428, opened 2025-01-28
mimxrt10xx/boards: Increase the flash CLK frequency.

robert-hh:
CLK is increased to the highest value defined in the data sheets of the respective flash device. The previous values of 60MHz and 30MHz reduce the speed of the APP. Tested with:

- arch_mix_1052
- imxrt1010_evk
- imxrt1015_evk
- imxrt1020_evk
- makerdiary_rt1011
- metro_m7_1011
- olimex_rt1010

### DIFF
--- a/ports/mimxrt10xx/boards/arch_mix_1052/flash_config.c
+++ b/ports/mimxrt10xx/boards/arch_mix_1052/flash_config.c
@@ -65,7 +65,7 @@ const flexspi_nor_config_t qspiflash_config = {
         .deviceModeArg = 0x40,
         .deviceType    = kFlexSpiDeviceType_SerialNOR,
         .sflashPadType = kSerialFlash_4Pads,
-        .serialClkFreq = kFlexSpiSerialClk_60MHz,
+        .serialClkFreq = kFlexSpiSerialClk_133MHz,
         .sflashA1Size  = FLASH_SIZE,
         .lookupTable =
         {
--- a/ports/mimxrt10xx/boards/imxrt1010_evk/flash_config.c
+++ b/ports/mimxrt10xx/boards/imxrt1010_evk/flash_config.c
@@ -64,7 +64,7 @@ const flexspi_nor_config_t qspiflash_config = {
         .deviceModeArg = 0x02,
         .deviceType    = kFlexSpiDeviceType_SerialNOR,
         .sflashPadType = kSerialFlash_4Pads,
-        .serialClkFreq = kFlexSpiSerialClk_60MHz,
+        .serialClkFreq = kFlexSpiSerialClk_100MHz,
         .sflashA1Size  = FLASH_SIZE,
         .lookupTable =
         {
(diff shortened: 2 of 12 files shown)


hathach:
Thanks you for your PR. Though I am off for TET (Lunar New Year) and won't be able to review this in 2 weeks. Happy New Year

ladyada:
fyi @tannewt

robert-hh:
@ladyada Thanks for noticing Scott. I'm not sure if the change I suggested is fine under all conditions. The flash chip in these boards would not support 100Mhz or 133Mhz with the simple read command (0x03). No problem with fast read. An alternative would be to change the flash frequency one the app is started, but I did not find a way to do so. Maybe not yet.

tannewt (approved):
I think this is ok because the LUT commands all do quad reads. This will speed up app startup. Thanks!

It is possible to change the frequency after startup. We do it here in CircuitPython: https://github.com/adafruit/circuitpython/blob/3236a0f200a3eb10594e6fec6f077fe0a2812d30/ports/mimxrt10xx/supervisor/internal_flash.c#L40-L85

robert-hh:
Thank you for both merging the PR and the hint for changing the frequency change. I have another question:
In the memconfig structure there are elements around `deviceModeCfgEnable`, which are configured to enable quad mode. It is configured to  send a `Write Status` command (hex 01), with argument 0x40; LUT entry 4. So far that's ok. But writing to the status register requires to send the `write Enable` command before. But that's missing. So I wonder how this enabling happens. Actually, the flash seem to run in quad mode. There is high frequency activity on all IO pins of the flash, even /WP and /HOLD.

tannewt:
I'm not sure how QE is set initially. I haven't looked at this in ages.

robert-hh:
I hooked up a logic analyzer to see what happens after reset
The ROM code executes:
a) A large block of data is read using command 0x03.
b) A write enable command 0x06 is sent
Then:
c) The startup command as defined by the arg of `deviceModeCfgEnable` is sent. So no magic. The ROM code itself send the Write Enable command.

robert-hh:
About changing the flash clock at runtime. I used the code from your link as template. As an element of pedantic, I add a mechanism to change the clock to any of the possible frequencies by modifying both the pfd_480 and the flexspi PODF dividers, using the values below which give the closest match.
```
typedef struct _ps_div_t {
    uint8_t pfd480_div;
    uint8_t podf_div;
} ps_div_t;

static ps_div_t div_table_mhz[] = {
    { 35, 8 }, // Index 0 is not valid, map it to 30 MHz
    { 35, 8 }, // 30 -> 30.85 MHz
    { 29, 6 }, // 50 -> 49.65 MHz
    { 18, 8 }, // 60 -> 60 MHz
    { 23, 5 }, // 75 -> 75.13 MHz
    { 18, 6 }, // 80 -> 80 MHz
    { 29, 3 }, // 100 -> 99.31 MHz
    { 13, 5 }, // 133 -> 132.92 MHz
    { 26, 2 }  // 166  -> 166.15 MHz
};
```

PR #430, opened 2025-02-04
mimxrt10xx/boards/olimex_rt1010/flash_config.c: Fix flash settings for the init sequence.

robert-hh:
## Description of Change

The settings for the initialization sequence of my last PR were not the right ones. Instead of the W25Q16 flash listed initially in the first Olimex documentation, the production boards now use an EN25Q16B-104HIP chip, which needs different arguments to enable quad mode. With the existing arguments, the flash may get write locked.

Tested with an Olimex RT1010Py board.

P.S.: In case the flash got write locked can then be unlocked using the flash-sfdp mode for uploading the UF2 bootloader.


tannewt (approved):
Thank you!

robert-hh:
Thank you for fast merging.

PR #439, opened 2025-03-02
Add support for the Daisy Seed board - stm32H750

snkYmkrct:
Added support for Electrosmith's  Daisy Seed board:  https://electro-smith.com/products/daisy-seed 

Application is loaded for this board, like the rest of stm32h7, on the external (Q)SPI flash. 
Code includes refactoring for the stm32h7 port to be able to support different external flash chips.


snkYmkrct:
> How? Will STM32H7 be supported eventually?

There already is support for the H7 and a [H750 board](https://github.com/adafruit/tinyuf2/tree/master/ports/stm32h7/boards/mini_stm32h750_v11) , this PR is just adding another board.

PR #440, opened 2025-03-12
Add Lilygo T-QT pro boards, N4R2 and N8

Neradoc:
There are two variants of the Lilygo T-QT Pro. I only own one but the difference is the ram/flash setup (N8 and N4R2 version). I used the VID/PID requested by @tyeth who didn't get to adding the board at the time.
It has a display, and no status LED.

PID: https://github.com/espressif/usb-pids/pull/80
Product page: https://lilygo.cc/products/t-qt-pro
Github page: https://github.com/Xinyuan-LilyGO/T-QT

Note that the non-pro are no longer available on Lilygo's site, but should be compatible with this.

### DIFF
--- a/ports/espressif/boards/lilygo_tqt_pro_nopsram/board.h
+++ b/ports/espressif/boards/lilygo_tqt_pro_nopsram/board.h
@@ -1,0 +1,86 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2020 Ha Thach (tinyusb.org) for Adafruit Industries
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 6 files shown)


ports/espressif/boards/lilygo_tqt_pro_nopsram/board.h
```diff
@@ -0,0 +1,86 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+#ifndef ADAFRUIT_QTPY_S3_H_
```
tyeth, line 25: Think this define guard could be updated to not match Adafruit qtpy.
Same in the other file.
Although I guess maybe only the correct file would be included based on board id / FQBN

Neradoc: Ah yes thanks !

tyeth (approved):

tyeth:
You mean you didn't add some custom sexy boot screen, I'm shocked ;)
Nice work on this @Neradoc, much appreciated!

PR #453, opened 2025-06-29
add UF2 family ID for ESP32-P4

lyusupov:
## Description of Change

Intension of this PR is to add UF2 family ID for ESP32-P4
https://github.com/microsoft/uf2/blob/master/utils/uf2families.json#L228

This action paves the way to further addition of ESP32-P4 target boards.


hathach (approved):
thank you

PR #473, opened 2026-04-10
Restore FlexSPI1 clock before app jump on RT1170

APIUM:
## Checklist

*By completing this PR sufficiently, you help us to review this Pull Request quicker and also help improve the quality of Release Notes*

*This checklist items that are not applicable to your PR can be deleted.*

-----------

## Description of Change

On testing a performance critical application I found I was getting slow LVGL performance compared to just flashing the hex files. Turns out the FlexSPI clock stays at 24MHz. This change addresses that, and also addresses your comment on https://github.com/adafruit/tinyuf2/pull/470 while I'm here to do the same change as early as possible too.

Tested on a custom 1176 board.

### DIFF
--- a/ports/mimxrt10xx/boards/imxrt1170_evk/board/clock_config.c
+++ b/ports/mimxrt10xx/boards/imxrt1170_evk/board/clock_config.c
@@ -160,10 +160,12 @@ void BOARD_BootClockRUN(void) {
     rootCfg.div = 1;
     CLOCK_SetRootClock(kCLOCK_Root_Cssys, &rootCfg);
 
-    /* Configure FLEXSPI1 using OSC_RC_48M_DIV2 */
-    rootCfg.mux = kCLOCK_FLEXSPI1_ClockRoot_MuxOscRc48MDiv2;
-    rootCfg.div = 1;
+    /* Configure FLEXSPI1 using OSC_RC_400M (100MHz) */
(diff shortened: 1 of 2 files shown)


ports/mimxrt10xx/boards.c
```diff
@@ -55,6 +55,18 @@ void board_init(void) {
   BOARD_InitBootPins();
   BOARD_BootClockRUN();
 
+  // Restore FlexSPI1 to high speed after PLL reconfiguration
+#if defined(MIMXRT117x_SERIES)
+  {
+    clock_root_config_t cfg = {0};
+    cfg.mux = kCLOCK_FLEXSPI1_ClockRoot_MuxOscRc400M;
+    cfg.div = 4;  // 400 MHz / 4 = 100 MHz
+    CLOCK_SetRootClock(kCLOCK_Root_Flexspi1, &cfg);
```
hathach: perfect, thank you. Quick question, should we move this into the BOARD_BootClockRUN() instead, maybe update the .mex file to generate the correct config ?

APIUM: Thanks for the point in the right direction here, I think I've done it right now? It's working on my board.

hathach (approved):
perfect, thank you. I forgot we currently doing manual clock config (intead of using mcxconfig for rt1170). But having it in the clock_config.c is right way to do

APIUM:
Thanks @hathach!

PR #478, opened 2026-06-22
update senseBox MCU Eye name and flashsize

PaulaScharf:
## Description of Change

We changed the board name from "senseBox Eye" to "senseBox MCU Eye". We also fixed the defined flash size for the Eye.


hathach (approved):
lgtm

